PR #226, opened 2026-06-25
fix(api): reject ambiguous username invite resolution

NateIsern:
### Motivation
- Inviting by username used a case-insensitive regex lookup that could match multiple case-variant accounts (e.g. `Alice` vs `alice`) and return an arbitrary document, allowing an attacker who registers a case-variant to be selected as the invite target.
- Usernames are stored trimmed but not lowercased and the signup uniqueness check was case-sensitive, so case-colliding usernames can coexist and make resolution ambiguous.

### Description
- Update `resolveUserByIdentifier` to treat case-insensitive username matches as ambiguous when more than one document is found and return `null` instead of selecting an arbitrary account, by using `User.find(...).limit(2)` and requiring exactly one match.
- Add `exactCaseInsensitiveUsernameRegex` (and export `escapeRegExp`) in `packages/api/src/utils/resolveUserIdentifier.ts` and reuse it to build the anchored, escaped, case-insensitive regex for username lookups.
- Use the shared `exactCaseInsensitiveUsernameRegex` in `packages/api/src/controllers/session.controller.ts` when checking username uniqueness during registration so new accounts that only differ by case are rejected.
- Add unit tests `packages/api/src/utils/resolveUserIdentifier.test.ts` that validate the regex escaping, unambiguous resolution, and rejection of ambiguous case-colliding matches.

### Testing
- Added unit tests for `resolveUserByIdentifier` (`resolveUserIdentifier.test.ts`) covering escaping, single-match resolution, and ambiguous two-match rejection; the tests are present but were not executed in this environment due to missing test runtime dependencies.
- Attempted to run package tests but `bun install` failed (npm registry returned `403`), and `jest` was not available, so automated tests could not be executed here and therefore their pass/fail status is unknown.
- Local static checks performed (`git diff --check`) showed no whitespace/patch errors prior to preparing the change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bbfd21fbc83238d5f86ad286e98da)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/226" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/controllers/session.controller.ts
+++ b/packages/api/src/controllers/session.controller.ts
@@ -27,6 +27,7 @@ import {
   clearAllRefreshCookies,
 } from '../services/refreshToken.service';
 import type { AuthRequest } from '../middleware/auth';
+import { exactCaseInsensitiveUsernameRegex } from '../utils/resolveUserIdentifier';
 
 /**
  * Constant-time dummy Argon2 hash used to keep login response timing
@@ -229,7 +230,7 @@ export class SessionController {
       }
 
       if (normalizedUsername) {
-        const existingUsername = await User.findOne({ username: normalizedUsername }).select('_id').lean();
+        const existingUsername = await User.findOne({ username: exactCaseInsensitiveUsernameRegex(normalizedUsername) }).select('_id').lean();
         if (existingUsername) {
           return res.status(409).json({ message: 'Username already taken' });
         }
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #231, opened 2026-06-25
fix(api): remove stale public CDN asset copies on visibility downgrade

NateIsern:
### Motivation

- A backfill copied legacy public objects into the `public/` CDN prefix without updating DB keys, allowing deterministic CDN URLs to remain reachable after a file's visibility is changed to `private`/`unlisted`.
- Visibility downgrade paths did not remove these backfilled `public/<key>` objects when the DB `storageKey` remained non-public, resulting in an access-control bypass for previously public media.

### Description

- Add `deleteBackfilledPublicCopy(key: string)` to `packages/api/src/services/assetService.ts` which deletes the `public/<key>` copy when present and logs failures.
- Invoke `deleteBackfilledPublicCopy` from `relocateObjectForVisibility` for non-public targets and from `deleteFile` for originals and variants so legacy CDN copies are removed on downgrade and permanent deletion.
- Add a regression unit test in `packages/api/src/services/__tests__/uploadCachedMediaStream.abort.test.ts` that verifies a legacy non-public DB key with a backfilled `public/<key>` copy is cleaned up when visibility is downgraded.

### Testing

- Added a focused unit test that exercises `relocateObjectForVisibility` for the legacy backfill case and asserts the `public/<key>` copy is deleted; the test file is `packages/api/src/services/__tests__/uploadCachedMediaStream.abort.test.ts`.
- Ran `git diff --check` successfully to validate formatting and diffs.
- Attempted to run package tests with `bun run --cwd packages/api test --runInBand <test>`, but dependency installation failed due to the environment returning HTTP 403 from the registry so the test run could not complete.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bb2864ab88323bdb17d3b56529a93)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/231" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/services/__tests__/uploadCachedMediaStream.abort.test.ts
+++ b/packages/api/src/services/__tests__/uploadCachedMediaStream.abort.test.ts
@@ -378,3 +378,35 @@ describe('uploadCachedMediaStream — abort cleanup', () => {
     source.destroy();
   });
 });
+
+describe('AssetService visibility relocation', () => {
+  it('deletes legacy backfilled public CDN copies when a non-public DB key is downgraded', async () => {
+    const existingKeys = new Set([
+      'content/2026/06/legacy-avatar.jpg',
+      'public/content/2026/06/legacy-avatar.jpg',
+    ]);
+    const deleteFile = jest.fn((key: string): Promise<void> => {
+      existingKeys.delete(key);
+      return Promise.resolve();
+    });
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #232, opened 2026-06-25
Prevent presigned PUT URL issuance for deduped live assets

NateIsern:
### Motivation

- The previous `initUpload` path returned a presigned S3 PUT URL for an existing deduplicated file record unconditionally, which allowed any authenticated user who knew a file's SHA-256 to obtain a write URL for the actual stored object and overwrite another user's asset.
- The goal is to preserve repair behavior for missing objects while removing the ability for non-owners to sign live deduplicated object keys.

### Description

- Change `AssetService.initUpload` to check the actual S3 object existence and only produce a repair presigned PUT URL when the object is missing and the requester is the file owner; otherwise return an empty `uploadUrl` for existing objects. (`packages/api/src/services/assetService.ts`)
- Keep the revive/repair logic for `deleted` records intact, but gate signing to owner-only missing-object repairs.
- Add regression tests that assert (1) a live existing object owned by another user does not get a PUT URL, and (2) the owner still receives a repair PUT URL when the object is missing. (`packages/api/src/services/__tests__/assetService.initUpload.test.ts`)

### Testing

- Added Jest unit tests in `packages/api/src/services/__tests__/assetService.initUpload.test.ts` to cover the regressed dedupe signing behavior; the tests are present in the branch but were not executed to completion in this environment.
- Attempted to run `bun run --filter @oxyhq/api test -- assetService.initUpload.test.ts`, which failed in this environment with `jest: command not found` so the tests could not be executed here.
- Attempted to run `bun run --filter @oxyhq/api build`, which failed before the API build due to TypeScript/tsconfig issues in `@oxyhq/contracts` in this runtime; CI should run with the proper toolchain where the new tests will execute as part of the package test suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bb6f7c148832385036f81ad7df6b6)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/232" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/services/__tests__/assetService.initUpload.test.ts
+++ b/packages/api/src/services/__tests__/assetService.initUpload.test.ts
@@ -1,0 +1,92 @@
+import { AssetService } from '../assetService';
+import { S3Service } from '../s3Service';
+
+jest.mock('../../utils/logger', () => ({
+  logger: { warn: jest.fn(), error: jest.fn(), info: jest.fn(), debug: jest.fn() },
+}));
+
+const mockFileFindOne = jest.fn();
+jest.mock('../../models/File', () => ({
+  File: { findOne: (...args: unknown[]) => mockFileFindOne(...args), findById: jest.fn() },
+  FileVisibility: {},
+}));
+
+jest.mock('../variantService', () => ({
+  VariantService: class {
+    constructor(_s3: unknown) { /* no-op */ }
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #247, opened 2026-06-25
Use Public Suffix List (tldts) for auth URL auto-detection

NateIsern:
### Motivation
- Prevent derivation of attacker-controllable IdP hosts by replacing a small hard-coded multi-part suffix list with a proper eTLD+1 computation using the Public Suffix List.

### Description
- Replace the ad-hoc `MULTIPART_TLDS` heuristic with `tldts.getDomain` in `registrableApex` so the registrable eTLD+1 (including private hosted suffixes) is computed correctly. 
- Keep `autoDetectAuthWebUrl` behavior (same-origin for `auth.*`, dev checks) but bail to `undefined` when a registrable apex cannot be determined. 
- Add `tldts` to `packages/core/package.json` and update `bun.lock` workspace metadata. 
- Update and extend `packages/core/src/utils/__tests__/fapiAutoDetect.test.ts` to add PSL/private-hosted suffix regression cases and adjust expectations accordingly. 
- Commit message: `fix(core): use PSL for auth URL detection`.

### Testing
- Ran `git diff --check` which succeeded without whitespace errors. 
- Attempted to run the core unit test file via `bun run test --runTestsByPath src/utils/__tests__/fapiAutoDetect.test.ts`, but the test run failed because `jest` is not available in this environment (`command not found`).
- Attempted to install `tldts` with `bun add tldts@^7.0.22 --cwd packages/core`, but the package registry fetch was blocked (HTTP 403), so dependency installation could not be completed here; `package.json` and `bun.lock` were updated to declare the dependency.
- Changes were committed locally (`fix(core): use PSL for auth URL detection`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bcec2d8b48323b8c2b3b974fddc45)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -319,6 +319,7 @@
         "invariant": "^2.2.4",
         "jwt-decode": "^4.0.0",
         "socket.io-client": "^4.8.1",
+        "tldts": "^7.0.22",
         "zod": "^3.25.64",
       },
       "devDependencies": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -104,6 +104,7 @@
     "invariant": "^2.2.4",
     "jwt-decode": "^4.0.0",
     "socket.io-client": "^4.8.1",
+    "tldts": "^7.0.22",
     "zod": "^3.25.64"
   },
   "peerDependencies": {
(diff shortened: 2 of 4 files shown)


NateIsern:

PR #248, opened 2026-06-25
fix(api): require ownership for email attachments

NateIsern:
### Motivation
- Prevent a privilege escalation where callers could attach and send File Manager files they do not own by passing accessible `fileId`s; view/access checks are insufficient for attachment/attachment-linking flows. 
- The outbound path downloads raw bytes by `fileId` with no caller context and `linkFile` mutates file metadata, so ownership must be enforced earlier to avoid cross-tenant data exposure or metadata tampering.

### Description
- Enforce ownership in the email controller: `resolveAttachmentInputs` now requires `file.ownerUserId.toString() === userId` before building the attachment snapshot or sending the message. (packages/api/src/controllers/email.controller.ts)
- Update the attachment schema comment to reflect that ownership is enforced by the controller rather than via a general access predicate. (packages/api/src/schemas/email.schemas.ts)
- Update tests to model ownership semantics, include `ownerUserId` in test fixtures, and assert non-owned active files are rejected before SMTP send or File Manager linking; also assert the old `canUserAccessFile` predicate is not relied upon in this path. (packages/api/src/routes/__tests__/email.send.attachments.test.ts)

### Testing
- Updated unit tests in `packages/api/src/routes/__tests__/email.send.attachments.test.ts` to cover the ownership rejection case; tests were adjusted to assert the new behavior.
- Attempted to run the package test (`bun run test -- packages/api/src/routes/__tests__/email.send.attachments.test.ts --runInBand`) but automated execution was blocked because `jest` is not installed in this environment, so the test suite was not executed here (command failed with `jest: command not found`).
- Changes were compiled into a commit for review and should be verified by CI which has project dependencies available; the intended unit tests should pass once run in the normal CI/workspace environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bd25a14988323a5919e1cc8b8649e)

### DIFF
--- a/packages/api/src/controllers/email.controller.ts
+++ b/packages/api/src/controllers/email.controller.ts
@@ -26,7 +26,7 @@ import type { RecipientInput, AttachmentInput } from '../schemas/email.schemas';
  * canonical IAttachment shape persisted on Message. Each fileId MUST:
  *   - exist
  *   - be in status 'active' (not trash/deleted)
- *   - be accessible to the requesting user (assetService.canUserAccessFile)
+ *   - be owned by the requesting user
  *
  * The IAttachment is built from the IFile so the Message subdocument carries
  * a stable snapshot of name/contentType/size at send time, regardless of any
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #249, opened 2026-06-25
fix(services): preserve global logout on web

NateIsern:
### Motivation
- Restore the intended "Sign out all sessions across all devices" behavior after a regression made the web `logoutAll` branch only clear local refresh cookies instead of revoking the user's sessions server-side.
- Ensure the Session Management UI and public `signOutAll` contract truly revoke remote sessions during an emergency global logout.

### Description
- Always call the bearer-backed global endpoint via `oxyServices.logoutAllSessions(activeSessionId)` before any web cookie cleanup in `useAuthOperations.logoutAll` (`packages/services/src/ui/context/hooks/useAuthOperations.ts`).
- On web, follow the global revocation with `oxyServices.logoutAllSessionsViaCookie()` and `clearActiveAuthuser()` so browser-local multi-account refresh-cookie slots are expired as well.
- Update the `useAuthOperations.logoutAll` unit tests to assert that the web path invokes both the global revocation and the cookie cleanup, and adjust the error test to reflect the new call order (`packages/services/__tests__/context/useAuthOperations.test.tsx`).
- Changes are limited to the services layer orchestration and tests; server-side endpoints (`/session/logout-all/:sessionId` and `/auth/logout`) remain unchanged and are used together to achieve correct semantics.

### Testing
- Ran `git diff --check` locally to validate whitespace/patch sanity and updated tests compile-wise in repository.
- Attempted to run the updated unit test (`useAuthOperations`), but `bun install` failed due to upstream registry `403` errors so Jest could not be executed in this environment; therefore the updated tests were not executed here.
- The change is covered by the modified unit test in `packages/services/__tests__/context/useAuthOperations.test.tsx` which asserts both the global revocation and cookie cleanup on web.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bd5acae5883239f4aad20211575b5)

### DIFF
--- a/packages/services/__tests__/context/useAuthOperations.test.tsx
+++ b/packages/services/__tests__/context/useAuthOperations.test.tsx
@@ -446,27 +446,22 @@ describe('useAuthOperations.logoutAll', () => {
   });
 
   // Under jest's jsdom env `isWebBrowser()` returns true, so `logoutAll`
-  // takes the web cookie-cleared branch. The bearer-protected
-  // `logoutAllSessions` is reserved for native and is NOT called on web —
-  // see the "Semantics intentionally diverge by platform" comment in
-  // `useAuthOperations.logoutAll` for why this is correct UX.
-  it('clears every device-local cookie + every session state on success (web)', async () => {
+  // must revoke the active user's sessions across devices first, then clear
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #250, opened 2026-06-25
fix(services): gate refresh-cookie restore on durable session

NateIsern:
### Motivation
- Prevent an attacker-controlled same-site origin or a cookie-tossing attack from exchanging a parent-domain `oxy_rt_*` refresh cookie for a JavaScript-visible bearer token during web cold boot.
- Harden the web cold-boot path so only origins with prior durable session proof can automatically restore full access tokens from the refresh-cookie exchange.

### Description
- Require origin-local durable session proof by reading the persisted `activeSessionId` from ready storage before calling the refresh-cookie fast-path in `restoreViaRefreshCookie` in `packages/services/src/ui/context/OxyContext.tsx`.
- Only accept a `refreshAllSessions` account whose returned `sessionId` matches the persisted active session id before planting the returned access token into the JS-visible HTTP client state.
- Update the dependency array of the `restoreViaRefreshCookie` callback to include the storage helpers used for the durable check.
- Add a regression test in `packages/services/__tests__/context/refreshCookieRestore.test.tsx` asserting that when no durable local session exists the code does not call `refreshAllSessions`, does not plant a token, and instead proceeds to the central SSO bounce; also update existing cookie-restore tests to seed the durable active-session id where appropriate.

### Testing
- `git diff --check` ran locally and reported no whitespace errors.
- Unit test run: attempted `bun run test -- refreshCookieRestore.test.tsx` but `jest` is not available in this ephemeral environment so the test could not be executed here.
- Type check: attempted `bun run typescript` but workspace compilation failed in this environment due to missing dev/runtime dependencies (e.g. `react`, `@oxyhq/core`, `react-native`) so a local `tsc` verification could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bd820f0788323b255b0d0240b9419)

### DIFF
--- a/packages/services/__tests__/context/refreshCookieRestore.test.tsx
+++ b/packages/services/__tests__/context/refreshCookieRestore.test.tsx
@@ -144,6 +144,8 @@ describe('Cold-boot restore via secure refresh cookies (multi-account)', () => {
 
   it('CASE single account: signs the user in from the refresh-all snapshot, plants the token, and persists session id + authuser', async () => {
     const stub = baseStub();
+    window.localStorage.setItem(ACTIVE_SESSION_KEY, COOKIE_SESSION_ID);
+
     stub.refreshAllSessions = jest.fn(async () => ({
       accounts: [
         {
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as not-correct. The threat model is false: refreshAllSessions → POST /auth/refresh-all only returns tokens for server-issued, server-rotated oxy_rt_* cookies — only the IdP can mint them, so a forged refresh cookie is impossible. Gating cookie-restore on a durable activeSessionId reintroduces exactly the stored-session dependency the feature was built to avoid, and breaks the legitimate first-party cold-boot cookie-restore step on *.oxy.so RPs that have valid Domain=oxy.so cookies but no prior localStorage session. (Where a durable session DOES exist, stored-session already wins earlier in cold boot.)

PR #251, opened 2026-06-25
fix(api): harden refresh cookie origin scope

NateIsern:
### Motivation
- Close a high-severity web logic vulnerability where a parent-domain refresh cookie (`oxy.so`) + permissive credentialed CORS allowed same-site subdomains to obtain a bearer `accessToken` via `POST /auth/refresh` without robust Origin/CSRF protections.

### Description
- Replace wildcard one-level subdomain trust with an explicit allowlist of known first-party app origins in `packages/api/src/config/allowedOrigins.ts`, removing the previous `ALLOWED_ORIGIN_PATTERNS` regex that matched any `*.oxy.so` subdomain.
- Make refresh cookies host-only by default by omitting the `Domain` attribute in `buildRefreshCookieOptions()` and only applying an emergency override when `REFRESH_COOKIE_DOMAIN` is explicitly set to the exact API host; update startup validation in `packages/api/src/config/env.ts` to reject unsafe values.
- When setting/clearing refresh cookies, explicitly expire any legacy parent-domain (`oxy.so`) cookie so previously issued parent-domain bearer cookies no longer leak to sibling subdomains; added helpers in `packages/api/src/services/refreshToken.service.ts` to clear legacy cookies.
- Update tests to assert the hardened policies: origin allowlist rejects `https://attacker.oxy.so`, `REFRESH_COOKIE_DOMAIN` validation behavior, and that cookie-clearing emits a `Domain=oxy.so` Max-Age=0 header to clean up legacy cookies.

### Testing
- Ran `git diff --check` successfully to ensure no whitespace/merge errors were introduced.
- Attempted `bun install` but dependency resolution failed due to external npm registry `403` responses, so workspace install could not complete.
- Attempted `bun run test` and `jest`-scoped runs but they were blocked because dependencies/installation did not complete; tests could not be executed locally here.
- Attempted `bun run api:build` but the build was blocked by existing TypeScript configuration deprecation/rootDir errors in the `@oxyhq/contracts` build step, so full compile verification could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bd87226008323bebbe29b2a3e8d8a)

### DIFF
--- a/packages/api/src/config/__tests__/env.test.ts
+++ b/packages/api/src/config/__tests__/env.test.ts
@@ -1,10 +1,10 @@
 /**
  * Environment configuration validation tests
  *
- * Focus: REFRESH_COOKIE_DOMAIN strict-hostname validation (MED-2). The value
- * is interpolated into a hand-built `Set-Cookie` header
- * (`appendLegacyRefreshCookieDeletion`), so anything beyond a bare hostname
- * (scheme, port, spaces, `;`, `,`, control chars) must fail fast at startup.
+ * Focus: REFRESH_COOKIE_DOMAIN strict validation (MED-2). Refresh cookies
+ * default to host-only scope. If the emergency Domain override is used, it must
+ * be exactly the API host so bearer-equivalent cookies never leak to sibling
+ * subdomain servers.
  */
 
 import {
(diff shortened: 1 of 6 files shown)


NateIsern:
✅ Landed in main via 33d7f09d. Refresh cookie is now host-only (api.oxy.so) with a legacy Domain=oxy.so clear, and wildcard *.oxy.so credentialed-CORS trust replaced by an explicit EXACT_ALLOWED_ORIGINS allowlist (includes os.oxy.so). NOTE: please eyeball the allowlist (posted in the task summary) so no live RP loses credentialed CORS.

PR #254, opened 2026-06-25
fix(auth): bind FedCM assertion client to request origin

NateIsern:
### Motivation
- Prevent FedCM client_id/origin spoofing by ensuring the IdP mints tokens only when the caller-supplied `client_id` is both approved and matches the actual browser `Origin`, closing an origin/nonce binding bypass.

### Description
- Require `/fedcm/assertion` to normalize and resolve `client_id` via the authoritative approved-client list and verify the normalized value equals the request `Origin`, rejecting the request when they differ and normalizing the signed `aud` to the approved origin (change in `packages/auth/server/index.ts`).
- Add regression coverage that rejects an approved-but-different `client_id` when it does not match the real request Origin and rejects unapproved `client_id` values before token minting (tests added/updated in `packages/auth/server/__tests__/fedcm.idp.test.ts`).
- Adjust the multi-domain assertion test to explicitly seed the expected approved origin in the test stub so the assertion path remains valid in that scenario.

### Testing
- Ran `git diff --check` which passed.
- Attempted `bun test packages/auth/server/__tests__/fedcm.idp.test.ts` but execution was blocked by unresolved workspace dependencies (`Cannot find module '@oxyhq/core'`), so the new tests could not be executed here.
- Attempted `bun install --frozen-lockfile` to install deps for full test runs but it failed due to registry access errors (multiple `403` responses), preventing local test execution in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3be07f2288832390814ba5aeb389f4)

### DIFF
--- a/packages/auth/server/__tests__/fedcm.idp.test.ts
+++ b/packages/auth/server/__tests__/fedcm.idp.test.ts
@@ -371,6 +371,49 @@ describe('POST /fedcm/assertion', () => {
     expect(payload.iss).toBe('https://auth.oxy.so');
   });
 
+  it('rejects when client_id is not the actual requesting Origin', async () => {
+    const approvedButDifferentOrigin = 'https://console.oxy.so';
+    stubbedApprovedClients = [RP_ORIGIN, approvedButDifferentOrigin];
+
+    const res = await app.request('/fedcm/assertion', {
+      method: 'POST',
+      headers: {
+        ...WEBIDENTITY,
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as duplicate of #253. Same /fedcm/assertion CORS+aud binding fix; #253 has the cleaner CORS posture (no credentialed CORS on error paths). #254's stronger approved-client check (Origin === resolveApprovedClientOrigin) is being folded into the landed version.

PR #255, opened 2026-06-25
fix(core): require delegated service scope intersection

NateIsern:
### Motivation
- Prevent delegated service requests from gaining scopes that were not explicitly granted by both the app and the per-user delegation, closing an authorization bypass where union semantics allowed escalation.

### Description
- Change `requireScope()` to require the requested scope to be present in both `req.serviceApp.scopes` and `req.serviceActingAs.scopes` when `req.serviceActingAs` is present, otherwise require the app scope for app-only requests (`packages/core/src/mixins/OxyServices.utility.ts`).
- Update the `requireScope()` documentation comment to describe the intersection semantics for delegated requests (`packages/core/src/mixins/OxyServices.utility.ts`).
- Add regression tests that assert delegated requests succeed only when both the app and the delegation carry the required scope and fail when only one side has it (`packages/core/src/mixins/__tests__/serviceAuth.test.ts`).

### Testing
- Ran `git diff --check` which reported no problems.
- Added unit tests for the `requireScope()` behavior but running `bun run test` locally was blocked because `jest` is not available in this environment and `bun install` failed due to the registry returning `403`, so tests could not be executed here.
- The new tests are included in `packages/core/src/mixins/__tests__/serviceAuth.test.ts` and should be run in CI or a developer environment with dependencies installed using `bun install` followed by the package test runner (`bun run test` / `jest` as appropriate).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3be13e2ad483239ef00edb327e805d)

### DIFF
--- a/packages/core/src/mixins/OxyServices.utility.ts
+++ b/packages/core/src/mixins/OxyServices.utility.ts
@@ -953,9 +953,9 @@ export function OxyServicesUtilityMixin<T extends typeof OxyServicesBase>(Base: 
      * Express.js middleware that enforces a specific service-token scope.
      *
      * Mount AFTER `auth()` / `serviceAuth()` — relies on `req.serviceApp` and
-     * (when delegation is in effect) `req.serviceActingAs.scopes`. The scope
-     * is granted if EITHER list contains it, mirroring the OAuth2 model where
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #258, opened 2026-06-25
fix: harden inbox email link handling

NateIsern:
### Motivation
- Close a critical sandbox-escape / popup-injection vulnerability introduced by allowing attacker-controlled email HTML to open unsanitized popups or trigger external app launches.  
- Ensure email links are only opened when they are clearly safe external URLs and when the navigation is user-initiated on native, preventing javascript: (mixed-case / whitespace-prefixed) and automatic navigations from being exploited.  

### Description
- Replaced permissive popup behavior by removing the injected `<base target="_blank">` and dropping `allow-popups` / `allow-popups-to-escape-sandbox` so email HTML remains confined to the iframe.  
- Added `getSafeExternalUrl()` which trims leading/trailing control/space characters, normalizes via `new URL(...)`, and only allows `http:`, `https:`, and `mailto:` schemes before any external open.  
- Updated the web iframe click handler to validate and normalize the clicked `href` with `getSafeExternalUrl()` and only call `window.open()` with the validated URL.  
- Hardened native WebView handling to require both a safe URL and an explicit user-click navigation (checks `navigationType === 'click'` / `isTopFrame`) before calling `Linking.openURL()`, preventing meta-refresh/iframe-driven external launches.  

### Testing
- Ran `git diff --check` to ensure no whitespace/patch issues and it passed.  
- Executed a small Node test exercising `getSafeExternalUrl()` against mixed-case, whitespace/control-prefixed, relative, javascript:, and valid URLs; the checks passed.  
- Attempted `bun run --filter inbox lint` but it could not run in this environment because the `expo` binary is not available (blocked).  
- Attempted `tsc --noEmit -p packages/inbox/tsconfig.json` (via the workspace) but full type-checking could not complete here due to missing workspace dependencies / Expo TS base config (blocked in this environment).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bef2f86f883238cb166bbef63ac8c)

### DIFF
--- a/packages/inbox/components/HtmlBody.tsx
+++ b/packages/inbox/components/HtmlBody.tsx
@@ -17,6 +17,30 @@ interface HtmlBodyProps {
   html: string;
 }
 
+const ALLOWED_EXTERNAL_LINK_PROTOCOLS = new Set(['http:', 'https:', 'mailto:']);
+const CONTROL_OR_SPACE_BOUNDARY = /^[\u0000-\u0020]+|[\u0000-\u0020]+$/g;
+
+function getSafeExternalUrl(href: string | null | undefined): string | null {
+  const trimmedHref = href?.replace(CONTROL_OR_SPACE_BOUNDARY, '');
+  if (!trimmedHref || trimmedHref.startsWith('#')) return null;
+
+  try {
+    const url = new URL(trimmedHref);
+    if (!ALLOWED_EXTERNAL_LINK_PROTOCOLS.has(url.protocol.toLowerCase())) {
+      return null;
+    }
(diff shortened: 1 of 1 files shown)


NateIsern:
✅ Landed in main via 33d7f09d, REWORKED. Web half applied as-is (tightened sandbox, protocol allowlist). Native half fixed: react-native-webview reports navigationType 'other'/null on Android, so the strict ==='click' guard would have blocked ALL Android email link taps — relaxed to accept click/other/null while keeping the protocol allowlist + top-frame check as the security boundary.

PR #259, opened 2026-06-25
fix(api): harden federation fetches against SSRF

NateIsern:
### Motivation
- Prevent unauthenticated SSRF introduced by resolving federated handles on the public `/profiles/username/:username` path, which previously allowed server-side fetches to attacker-controlled WebFinger/actor/avatar URLs.
- Ensure federated outbound fetches cannot follow redirects into private IP ranges, localhost, or non-HTTPS schemes and that fetched actor IDs are validated before use.

### Description
- Add URL/IP/DNS validation helpers: `assertPublicHttpsUrl`, `isPrivateAddress`, and a `MAX_FEDERATION_REDIRECTS` limit to ensure only public `https:` destinations are followed and redirects are rechecked.
- Add `safeFederationFetch` wrapper that resolves hostnames via `dns.lookup`, blocks private/reserved IPs and localhost-style names, enforces `https:`, rejects credentials in URLs, and re-validates `Location` on redirects.
- Route all federation outbound IO through the guard: replaced direct `fetch()` calls for WebFinger, signed actor fetch (`signedFetch`), and avatar downloads with `safeFederationFetch` to harden WebFinger→actor→avatar stages.
- Validate returned actor `id` values via `assertPublicHttpsUrl` before extracting `hostname` and persisting `actorUri` metadata, and require `actor.id` to be a string to avoid unsafe URIs being accepted.

### Testing
- Ran `git diff --check` to verify no accidental whitespace/errors; it passed.
- Ran a Bun-based transpile smoke check with `bun -e` to ensure the edited TS parses; the transpiler returned `ok`.
- Attempted package build with `bun run --cwd packages/api build`, which was blocked by unrelated TypeScript config errors in `@oxyhq/contracts` (TypeScript/tsconfig issues), so full type-checked build could not be completed here.
- `bun install --frozen-lockfile` failed due to upstream registry `403` errors, preventing running the full test suite; integration/unit tests were therefore not executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bf98dbc848323a80ae32fa75334aa)

### DIFF
--- a/packages/api/src/services/federation.service.ts
+++ b/packages/api/src/services/federation.service.ts
@@ -7,6 +7,8 @@
  */
 
 import crypto from 'crypto';
+import dns from 'dns/promises';
+import net from 'net';
 import mongoose from 'mongoose';
 import User, { IUser } from '../models/User';
 import { AssetService } from './assetService';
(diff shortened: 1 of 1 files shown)


NateIsern:
✅ Superseded by the consolidated federation SSRF fix landed in main via 33d7f09d. Rather than hand-rolling SSRF checks (which carried a DNS-rebind TOCTOU), all three federation fetch sites now route through @oxyhq/core/server safeFetch (DNS-pinned, private-IP denylist, redirect-revalidating), https-only, plus streaming byte caps. Your https-only + actor.id host check were incorporated.

PR #260, opened 2026-06-25
fix(api): harden federated avatar downloads against SSRF

NateIsern:
### Motivation
- Prevent server-side request forgery and data exfiltration caused by downloading attacker-controlled avatar URLs from privileged service-authenticated callers.

### Description
- Add a safe-fetch wrapper (`fetchSafeAvatarUrl`) and URL assertion (`assertSafeAvatarUrl`) that parse and validate `avatar` URLs, enforcing HTTPS-only, no embedded credentials, no custom ports, and rejecting `localhost` and literal/resolved private or reserved IP addresses.
- Follow redirects manually with a limit (`MAX_AVATAR_REDIRECTS`) and validate each redirect target before following to avoid redirect-based bypasses.
- Introduce a pre-buffer `Content-Length` guard and a shared `MAX_AVATAR_DOWNLOAD_BYTES` (5MB) constant to reject oversized responses before reading the full body while keeping the existing post-read size check.
- Replace the direct `fetch(avatarUrl, ...)` call in `downloadAndStoreAvatar` with the safe-fetch path, preserving conditional (`ETag`/`Last-Modified`) handling and existing upload logic, and add unit tests that assert unsafe URLs are blocked and oversized responses are rejected.

### Testing
- Added Jest unit tests in `packages/api/src/services/__tests__/federation.service.test.ts` covering unsafe URL blocking and oversized `Content-Length` handling; these tests mock DNS resolution and `fetch` behavior.
- Attempted to run the package tests with `bun run test -- federation.service.test.ts` but execution was blocked because `jest` is not available in the ephemeral environment (dev dependencies are not installed), so tests were not executed here.
- Attempted `bun install --frozen-lockfile` to install deps but it failed due to npm registry HTTP 403 responses in this environment, preventing local test execution; a static diff/check was performed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bfc020d2483238f88545b458daf87)

### DIFF
--- a/packages/api/src/services/__tests__/federation.service.test.ts
+++ b/packages/api/src/services/__tests__/federation.service.test.ts
@@ -24,6 +24,7 @@ const mockCacheInvalidate = jest.fn();
 const mockAssetFileContentExists = jest.fn();
 const mockAssetUploadFileDirect = jest.fn();
 const mockAssetDeleteFile = jest.fn();
+const mockDnsLookup = jest.fn();
 
 process.env.AWS_ACCESS_KEY_ID ||= 'test-access-key';
 process.env.AWS_SECRET_ACCESS_KEY ||= 'test-secret-key';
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as duplicate of #264 (and superseded). #260 protects only the avatar-download fetch site — a strict subset of the three federation fetch sites. The federation SSRF fix is being landed consolidated, routed through @oxyhq/core/server safeFetch (DNS-pinned, redirect-revalidating) instead of a hand-rolled check.

PR #261, opened 2026-06-25
fix: pin Docker installs to committed bun.lock to prevent live dependency resolution

NateIsern:
### Motivation
- Prevent production Docker builds from resolving mutable semver ranges by restoring use of the committed `bun.lock`, closing a supply-chain window where rebuilt images could pull unreviewed package versions.

### Description
- Updated `Dockerfile` builder stage to `COPY package.json bun.lock ./` and run `bun install --frozen-lockfile` so build deps are installed from the reviewed lockfile. 
- Updated `Dockerfile` production stage to `COPY package.json bun.lock ./` and run `bun install --production --frozen-lockfile` so runtime deps are also pinned.
- Added `COPY` entries for each workspace `package.json` in the Dockerfile so the full workspace lockfile is valid for a frozen install.
- Updated `.dockerignore` to exclude workspace source trees while allowing their `package.json` manifests into the build context so Docker can perform a full-workspace frozen Bun install.

### Testing
- Ran `git diff --check` and a local Node verification script that asserts every workspace `package.json` is copied and that `bun install --frozen-lockfile` is present; these checks passed.
- Attempted a temporary `bun install --frozen-lockfile --production` in a simulated context which reached dependency fetches but failed with registry `403` in this environment (network restrictions), indicating no lockfile mutation error but preventing a full install verification here.
- Attempted `docker build` in the environment but Docker is not available, so an image build could not be executed; CI/build should validate the frozen install on the runner where Docker and registry access are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bff34175883238a1cbe2195b3ca21)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -49,16 +49,24 @@ dkim-private.pem
 **/.eslintcache
 
 # Other apps in the monorepo that the API image does not build.
-# The Dockerfile only copies packages/core and packages/api, but
-# excluding the rest keeps the build context lean.
-packages/accounts
-packages/auth
-packages/auth-sdk
-packages/console
-packages/inbox
-packages/services
-packages/test-app-expo
-packages/test-app-vite
+# Keep their sources out of the image context, but allow package manifests so
+# Docker can run a full-workspace frozen Bun install against the committed lock.
+packages/accounts/*
+!packages/accounts/package.json
+packages/auth/*
+!packages/auth/package.json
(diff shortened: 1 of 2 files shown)


NateIsern:
✅ Landed in main via 33d7f09d (winner over #262 — also fixes .dockerignore so the frozen install actually builds). #262 closed as dup.

PR #266, opened 2026-06-25
fix(api): harden unsubscribe SSRF validation

NateIsern:
### Motivation
- The unsubscribe flow performed server-side `fetch()` to attacker-controlled `List-Unsubscribe` URLs while only checking the literal hostname against regexes, which is bypassable via DNS and redirects and creates a blind SSRF primitive. 
- The server followed redirects by default and did not re-validate redirect targets, so an attacker-controlled HTTPS host could redirect to internal/private addresses. 
- Timeouts were not guaranteed to be cleared on all failure paths, risking resource leaks during outbound requests.

### Description
- Hardened `packages/api/src/services/email.service.ts` by importing `lookup` from `dns/promises` and `net` and adding an `isPrivateIpAddress` helper to classify IPs. 
- Made `validateUnsubscribeUrl` asynchronous and now DNS-resolves the unsubscribe host, rejecting hosts that resolve to localhost, link-local, private, shared, benchmark, multicast, or otherwise non-public addresses. 
- Disabled automatic redirect following for RFC8058 POST and HTTP GET unsubscribe fetches (`redirect: 'manual'`) and explicitly reject 3xx responses instead of following them, and ensured timeout cleanup is done in `finally` blocks. 

### Testing
- Ran `git diff --check` which passed locally. 
- Attempted `bun run --filter @oxyhq/api build` and `bunx tsc -p packages/api/tsconfig.json --noEmit --ignoreDeprecations 6.0` but these were blocked by existing TypeScript config deprecation/rootDir errors and missing installed dependencies in the environment. 
- Attempted `bun install` and `bun --check` type checks but they failed due to registry `403` responses and unavailable `node_modules`, so full compile/test could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c114349d48323bbe902bd25b26d95)

### DIFF
--- a/packages/api/src/services/email.service.ts
+++ b/packages/api/src/services/email.service.ts
@@ -8,6 +8,8 @@
  */
 
 import mongoose, { type SortOrder } from 'mongoose';
+import { lookup } from 'dns/promises';
+import net from 'net';
 import { Mailbox, IMailbox } from '../models/Mailbox';
 import { Message, IMessage, IEmailAddress, IAttachment } from '../models/Message';
 import { Label } from '../models/Label';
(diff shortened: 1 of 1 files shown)


NateIsern:
✅ Landed in main via 33d7f09d, REWORKED. The unsubscribe fetch now routes through @oxyhq/core/server safeFetch (DNS-pinned, https-only, body destroyed unread) instead of the hand-rolled DNS check + redirect:manual, which carried the same DNS-rebind TOCTOU.

PR #267, opened 2026-06-25
Disable public SMTP in API container

NateIsern:
### Motivation
- The unified deployment exposed public SMTP (`25`/`587`) from the same container that runs the HTTP API and forced `SMTP_ENABLED=true`, which co-locates unauthenticated SMTP processing with the API and creates a network-accessible resource-exhaustion/DoS risk. 
- Production inbound mail is intended to be Cloudflare Email Routing → Worker → `POST /email/inbound`, so the default sample config should not enable on-box SMTP that can impact API availability.

### Description
- Set `SMTP_ENABLED=false` in `.env.example` and added guidance that self-hosted SMTP should run in a dedicated mail worker/container. 
- Updated `Dockerfile` docs and removed public SMTP port exposure so the image documents/`EXPOSE`s only `8080` (the API), leaving the API entrypoint unchanged. 
- Updated `docker-compose.yml` to remove `ports` mappings for `25:25` and `587:587`, set `SMTP_ENABLED=false` in the `api` service environment, and removed the on-container SMTP cert mount / direct `rspamd` dependency for the API service. 
- Retained the SMTP implementation gated by `SMTP_ENABLED` in code so operators can opt into a separate mail worker, but the sample API container no longer enables or exposes SMTP by default.

### Testing
- Ran `git diff --check` which completed without whitespace errors. 
- Ran pattern searches (`rg`) to verify that `SMTP_ENABLED=true` and public SMTP port mappings are no longer present in `.env.example`, `Dockerfile`, and `docker-compose.yml`, which succeeded. 
- Attempted `docker compose config` to validate the compose file, but Docker is not available in this environment so compose validation could not be executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c3009c4c88323a78e1b507bfba6b8)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,7 +19,9 @@ AWS_ACCESS_KEY_ID=your-key
 AWS_SECRET_ACCESS_KEY=your-secret
 
 # ─── Email ────────────────────────────────────────────────────────
-SMTP_ENABLED=true
+# SMTP is disabled by default so public mail handling does not share the API process.
+# For self-hosted SMTP deployments, run a dedicated mail worker/container instead.
+SMTP_ENABLED=false
 EMAIL_DOMAIN=oxy.so
 
 # Cloudflare Email Routing inbound webhook secret
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
 ##
-## Dockerfile for the Oxy API Server (with built-in email worker)
+## Dockerfile for the Oxy API Server
 ##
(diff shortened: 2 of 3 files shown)


NateIsern:
✅ Landed in main via 33d7f09d. Disables the legacy public SMTP listener in the API container (prod inbound uses Cloudflare Email Routing → Worker → /email/inbound; outbound uses the SES relay — both untouched).

PR #268, opened 2026-06-25
fix(core): require validated sessions for Socket.IO auth

NateIsern:
### Motivation
- Close a critical authentication bypass in the Socket.IO middleware that trusted decoded JWT claims without cryptographic or server-side verification, which allowed attacker-supplied tokens to impersonate arbitrary users.

### Description
- Hardened `authSocket()` in `packages/core/src/mixins/OxyServices.utility.ts` to require `sessionId` in the token payload and to call `validateSession()` before authenticating the socket.  
- Compare the server-validated session user to the claimed `userId` from the decoded token and reject the connection on mismatch.  
- Added a helper `getUserIdentityId` to safely extract normalized `id`/`_id` from the session `user` object.  
- Added regression tests at `packages/core/src/__tests__/authSocket.test.ts` covering: missing `sessionId`, decoded-vs-validated user mismatch, and the valid matched-session path.

### Testing
- Added unit tests for `authSocket()` (new test file `packages/core/src/__tests__/authSocket.test.ts`), but executing them in this environment was blocked because dependency resolution failed.  
- Attempted `bun install` (dependency fetch failed with HTTP 403 in this environment) and `bun run test` / typecheck runs therefore did not complete.  
- Ran local static checks (`git diff --check`) and validated the code diff and new test file were created as expected.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c2ddabd888323ade34c26cd0dda4f)

### DIFF
--- a/packages/core/src/__tests__/authSocket.test.ts
+++ b/packages/core/src/__tests__/authSocket.test.ts
@@ -1,0 +1,96 @@
+import { OxyServices } from '../OxyServices';
+
+function jsonResponse(data: unknown): Response {
+  return new Response(JSON.stringify({ data }), {
+    status: 200,
+    headers: { 'content-type': 'application/json' },
+  });
+}
+
+function createJwt(payload: Record<string, unknown>): string {
+  const encode = (value: unknown): string => Buffer.from(JSON.stringify(value)).toString('base64url');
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #273, opened 2026-06-25
fix(api): harden public email proxy redirects

NateIsern:
### Motivation
- Prevent unauthenticated SSRF via the public `/email/proxy` endpoint by strengthening URL validation and preventing attacker-controlled redirects from reaching internal/private hosts.
- The previous implementation only validated the original hostname with regexes and used `redirect: 'follow'`, allowing a malicious initial URL to redirect the server to localhost/RFC1918/metadata endpoints.

### Description
- Replaced the regex-only validation with `validateProxyUrl` that performs DNS `lookup` and rejects hosts or resolved addresses that map to loopback, link-local, multicast, or private ranges, and disallows `localhost`/`.local` names; this function is now async and uses `net.isIP` for address checks. (`packages/api/src/controllers/emailProxy.controller.ts`)
- Switched fetch behavior from automatic redirect following to manual redirect handling with a `MAX_REDIRECTS` cap and revalidation of each redirect target via `validateProxyUrl` before proceeding to the next hop. (`packages/api/src/controllers/emailProxy.controller.ts`)
- Preserved existing safeguards including content-type checks, font-extension fallback, size limits, tracking-pixel blocking, caching, and response headers, while adapting logic to reference the final validated `currentUrl` after redirects. (`packages/api/src/controllers/emailProxy.controller.ts`)
- Added regression tests that exercise direct private URLs, hostnames resolving to private addresses, blocked private redirects, and safe redirects that are revalidated then proxied. (`packages/api/src/controllers/__tests__/emailProxy.controller.test.ts`)

### Testing
- `git diff --check` and local repository sanity checks were run and passed for the applied changes. (succeeded)
- Added unit tests for `proxyResource` covering private-host rejection and redirect revalidation, but running the test suite was blocked because `bun install` failed due to npm registry access returning HTTP 403 in this environment, so `jest` could not be executed. (blocked)
- `bun install` was attempted to prepare test dependencies but failed due to registry access errors, preventing automated test execution in this environment. (blocked)

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c0d0074a0832398d7c53a73c35545)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/273" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #275, opened 2026-06-25
fix(api): harden email proxy SSRF checks

NateIsern:
### Motivation
- The email proxy previously validated only the literal hostname with regexes and used `fetch(..., redirect: 'follow')`, which allowed authenticated users to trigger server-side requests that could follow redirects or resolve to private/internal IPs (SSRF). 
- The content-type check occurred after the fetch completed, so blind SSRF and internal probing remained possible even when responses were later rejected. 
- The goal is to prevent requests from reaching non-public/internal addresses and to validate every redirect target while preserving existing proxy features (caching, tracking protection, size limits). 

### Description
- Added DNS and socket-level validation helpers (`normalizeHostname`, `isPrivateHost`, `isPublicIp`, `assertPublicDnsTarget`) and expanded the private/reserved IP pattern list to cover CGNAT, TEST-NET, multicast/reserved and IPv6 cases. 
- Replaced automatic redirect-following `fetch` usage with a manual `fetchPublicResource` helper that performs a DNS resolution check before connecting, enforces the same public-IP check in the `lookup` used by the outbound socket, and revalidates each redirect target with a `MAX_REDIRECTS` limit. 
- Updated `proxyResource` to call `fetchPublicResource`, use the fetch helper's `finalUrl` for extension/mime heuristics, and preserve existing behavior for content-type validation, tracking-pixel blocking, size limits, caching, and transparent-gif fallbacks. 
- Minimal API surface change: only `packages/api/src/controllers/emailProxy.controller.ts` was modified and the route/behavior callers remain unchanged. 

### Testing
- Ran `git diff --check` to validate formatting and found no whitespace issues. 
- Attempted to run package tests via `bun run --cwd packages/api test -- emailProxy`, but the test runner could not be executed because dependencies could not be installed in this environment (`bun install` failed with npm registry 403), so automated tests did not run. 
- Attempted TypeScript type-check with `npx tsc -p packages/api/tsconfig.json --noEmit --ignoreDeprecations 6.0`, but type-check failed due to missing runtime/dev dependencies and type packages in the isolated environment, so CI-style validation could not be completed locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c1d409d7c8323854763079ff14a5a)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/275" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/controllers/emailProxy.controller.ts
+++ b/packages/api/src/controllers/emailProxy.controller.ts
@@ -6,6 +6,10 @@
  */
 
 import { Request, Response } from 'express';
+import dns from 'dns';
+import http from 'http';
+import https from 'https';
+import net from 'net';
 import { BadRequestError } from '../utils/error';
 import { logger } from '../utils/logger';
 import crypto from 'crypto';
@@ -14,6 +18,7 @@ import crypto from 'crypto';
 
 const MAX_IMAGE_SIZE = 10 * 1024 * 1024; // 10 MB
 const FETCH_TIMEOUT = 10000; // 10 seconds
+const MAX_REDIRECTS = 5;
 const TRACKING_PIXEL_THRESHOLD = 100; // bytes
 const FONT_EXTENSIONS = /\.(ttf|otf|woff2?|eot)(\?|$)/i;
 const FONT_MIME: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


NateIsern:
Closing as duplicate of #273 (same emailProxy.controller SSRF rewrite). #273 is the cleaner base; #275's custom http/https `lookup` re-introduces the documented Bun `{all:true}` lookup-array footgun and ships no tests. The fix is being landed reworked onto @oxyhq/core/server safeFetch.

PR #282, opened 2026-06-25
fix(api): prevent bulk unfollow counter races

NateIsern:
### Motivation
- Prevent an integrity race where concurrent `bulkUnfollow` calls could over-decrement `_count.followers` and `_count.following` by basing counter updates on a pre-delete read instead of actual deletions.
- Ensure follower/following counters remain consistent with the true follow graph under concurrent requests.
- Maintain idempotent, unfollow-only semantics while preserving per-target result accuracy returned to callers.

### Description
- Replace the single `deleteMany(...)` with per-target atomic `findOneAndDelete(...)` calls and derive the subsequent counter decrements from the set of documents actually deleted in `packages/api/src/services/user.service.ts`.
- Update the returned per-target `wasFollowing` entries and the `unfollowedCount` to reflect only those targets that were actually removed instead of the pre-read candidate set.
- Add unit tests exercising a partial-race deletion and an all-raced-away case to validate that counters are decremented only for documents this request removed at `packages/api/src/services/__tests__/user.service.bulkUnfollow.test.ts`.

### Testing
- Added unit tests covering raced deletions and no-op counter updates when another request already removed the follow, but test execution was not possible in this environment due to missing `jest` and unavailable package installs.
- Performed a basic diff/consistency check (`git diff --check`) which reported no whitespace or diff issues.
- Attempted dependency installation (`bun install`) and running the test (`bun run test -- user.service.bulkUnfollow.test.ts`) but both were blocked in CI-like environment because the npm registry downloads returned `403` errors, so automated test execution could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db2f7c83239fb31dcf5a285513)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/282" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/services/__tests__/user.service.bulkUnfollow.test.ts
+++ b/packages/api/src/services/__tests__/user.service.bulkUnfollow.test.ts
@@ -1,0 +1,117 @@
+import { Types } from 'mongoose';
+
+const mockFollowFindLean = jest.fn();
+const mockFollowFindSelect = jest.fn(() => ({ lean: mockFollowFindLean }));
+const mockFollowFind = jest.fn(() => ({ select: mockFollowFindSelect }));
+const mockFollowFindOneAndDeleteLean = jest.fn();
+const mockFollowFindOneAndDeleteSelect = jest.fn(() => ({ lean: mockFollowFindOneAndDeleteLean }));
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #283, opened 2026-06-25
fix(api): prevent revival/claiming of deleted assets by SHA

NateIsern:
### Motivation
- Prevent an authorization/ownership bug where a caller-supplied SHA could match a deleted `File` record and cause that deleted record to be reactivated and transferred to the caller.
- Restore the original ownership/deletion boundary so tombstoned records remain isolated from client-supplied hash-based deduplication.

### Description
- Scope SHA de-duplication lookups to non-deleted records by introducing `findActiveFileBySha(sha256)` and replacing all `findFileBySha` usages in `packages/api/src/services/assetService.ts` with the active-only lookup so `status: 'deleted'` records are excluded from reuse.
- Remove deleted-record revival logic from `initUpload`, direct upload, streamed upload, and duplicate-race handling so deleted files are no longer reactivated or reassigned to callers.
- Change the `File` schema (`packages/api/src/models/File.ts`) so `sha256` uniqueness is enforced only for non-deleted records via a partial unique index (`partialFilterExpression: { status: { $in: ['active','trash'] } }`), and remove the unconditional `unique: true` on the field.
- Update the unit test `packages/api/src/services/__tests__/uploadCachedMediaStream.abort.test.ts` to assert that SHA de-duplication queries include `status: { $ne: 'deleted' }` instead of expecting deleted-record revival behavior.

### Testing
- Ran `git diff --check` to validate patch formatting and simple diffs, which passed.
- Attempted dependency install with `bun install --frozen-lockfile` but it was blocked by external npm registry 403 responses so full test execution could not run in this environment.
- Attempted to run the package test subset (`cd packages/api && bun run test -- <test-file>`), but `jest` was not available because dependencies failed to install, so unit tests could not be executed here; tests were updated to reflect the new non-revival behavior and are expected to pass once dependencies are installed and `jest` is available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db87508323ac3a6e3b1192107e)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/283" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/api/src/models/File.ts
+++ b/packages/api/src/models/File.ts
@@ -89,7 +89,6 @@ const FileSchema = new Schema<IFile>({
   sha256: { 
     type: String, 
     required: true, 
-    unique: true,
     index: true 
   },
   size: { type: Number, required: true },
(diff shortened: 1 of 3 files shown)


NateIsern:
Superseded by #414. The assetService dedup-security logic (`findActiveFileBySha` + removal of deleted-asset revival) is already in main (commit c54d24d4), and the File.ts sha256 live-only partial unique index plus the essential legacy `sha256_1` index-drop migration landed via #414 (`ensureFileSha256LiveUniqueIndex`). #283 lacked that migration, so closing in favor of the complete fix.

PR #292, opened 2026-06-25
fix(core): bind SSO code exchange to state

NateIsern:
### Motivation
- The central SSO helpers exposed a flow where an opaque SSO `code` could be exchanged and installed without verifying the returned `state`, enabling login-CSRF / session-fixation attacks. 
- The change ensures the client-side exchange path enforces the same per-origin CSRF `state` check that other SSO flows already performed, preventing attacker-supplied codes from replacing a victim's session.

### Description
- Threaded the SSO `state` through the return-consumption path so the token-committing call receives both the `code` and `state` (updated `consumeSsoReturn` to call `exchangeSsoCode(code, state)`).
- Added a browser-side guard in the client `exchangeSsoCode` that checks any stored per-origin SSO state (via `sessionStorage`) and rejects the exchange when the returned state does not match, preventing POST/fetch and token installation on mismatch.
- Updated public interface/type to accept an optional `state` parameter on `exchangeSsoCode`, added a small helper `getStoredSsoStateForCurrentOrigin`, and planted the state-check before the `/sso/exchange` request.
- Adjusted and added unit-test assertions across `@oxyhq/core`, `@oxyhq/services`, and `@oxyhq/auth-sdk` tests to assert the state-bound exchange call and to add an explicit test that a mismatched stored state blocks the exchange.

### Testing
- Ran `git diff --check` which reported no whitespace/errors and the commit was created successfully. 
- Attempted to run the core SSO unit tests with `bun run --cwd packages/core test -- ssoReturn.test.ts sso.test.ts` but automated execution in this environment failed because test dependencies / `jest` are not installed (`jest: command not found`), so tests could not be executed here; unit tests were updated in the change-set to cover the new behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8b7bc8323b682580537b60a3e)

### DIFF
--- a/packages/auth-sdk/__tests__/WebOxyProvider.coldBoot.test.tsx
+++ b/packages/auth-sdk/__tests__/WebOxyProvider.coldBoot.test.tsx
@@ -49,7 +49,7 @@ interface CoreStubs {
   handleRedirectCallback: jest.Mock<SessionLoginResponse | null, []>;
   isFedCMSupported: jest.Mock<boolean, []>;
   silentSignInWithFedCM: jest.Mock<Promise<SessionLoginResponse | null>, []>;
-  exchangeSsoCode: jest.Mock<Promise<SessionLoginResponse>, [string]>;
+  exchangeSsoCode: jest.Mock<Promise<SessionLoginResponse>, [string, string?]>;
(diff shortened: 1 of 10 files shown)


NateIsern:

PR #293, opened 2026-06-25
security: disable IdP auto-detection for shared-hosting suffixes

NateIsern:
### Motivation
- Close a session-fixation trust boundary where `autoDetectAuthWebUrl()` could map tenant hosts like `victim.pages.dev` → `https://auth.pages.dev`, allowing an attacker-controlled sibling IdP to silently plant sessions. 
- Ensure the SDK fails closed for multi-tenant/shared hosting suffixes and requires an explicit `authWebUrl` when auto-detection would be unsafe.

### Description
- Added a `SHARED_HOSTING_SUFFIXES` denylist to `packages/core/src/utils/fapiAutoDetect.ts` and made `registrableApex()` return `null` for those suffixes so `autoDetectAuthWebUrl()` will not derive `auth.<suffix>` for shared-hosting domains.
- Updated `packages/services/src/ui/context/OxyContext.tsx` to reuse the hardened `registrableApex()` in the same-site IdP check so visibility-driven `/auth/session-check` logout decisions also fail closed on shared/public suffixes.
- Exported the new `SHARED_HOSTING_SUFFIXES` from `@oxyhq/core` and added regression tests in `packages/core/src/utils/__tests__/fapiAutoDetect.test.ts` covering representative shared-hosting suffixes (e.g. `pages.dev`, `github.io`, `appspot.com`, `vercel.app`).
- Modified the code-paths that previously used a naive "last two labels" apex heuristic to use the strengthened logic.

### Testing
- Ran `bun test packages/core/src/utils/__tests__/fapiAutoDetect.test.ts` and all tests in that file passed.
- Ran `git diff --check` to confirm no whitespace/format errors and observed no issues.
- Could not run the full workspace test suite because dependency installation failed in this environment (`bun install` returned registry 403 and `jest` was unavailable), so broader package-level Jest runs were not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dbf1b48323a0c086761047791d)

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -462,7 +462,7 @@ export type { QuickAccount, DisplayNameUserShape } from './utils/accountUtils';
 // ---------------------------------------------------------------------------
 // Cross-domain SSO infrastructure
 // ---------------------------------------------------------------------------
-export { autoDetectAuthWebUrl, registrableApex, MULTIPART_TLDS } from './utils/fapiAutoDetect';
+export { autoDetectAuthWebUrl, registrableApex, MULTIPART_TLDS, SHARED_HOSTING_SUFFIXES } from './utils/fapiAutoDetect';
 
 // Central cross-domain SSO (opaque single-use code bounce via auth.oxy.so)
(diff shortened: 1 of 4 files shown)


NateIsern:

PR #295, opened 2026-06-25
fix(api): preserve client device fingerprints

NateIsern:
### Motivation
- A client-side change started sending a 64-char SHA-256 hex string as `deviceFingerprint`, but the server fingerprint pipeline expected a structured object which produced the same empty-hash for all string inputs and collapsed distinct device logins into one device/session for the same user.
- The goal is to accept and preserve legitimate client-submitted fingerprint strings while keeping existing structured-object fingerprint behavior intact so device/session separation and security monitoring remain correct.

### Description
- Add a union input type `DeviceFingerprintInput = DeviceFingerprint | string` and a `CLIENT_FINGERPRINT_HEX_RE` check so server helpers accept either a structured object or a client-provided string and normalize 64-char hex fingerprints to lowercase.
- Modify `generateDeviceFingerprint` to return the client-provided 64-char hex as-is (normalized), otherwise hash the provided string input or continue hashing the structured object as before so behavior is backwards-compatible.
- Update session types so `SessionCreateOptions.deviceFingerprint` reflects the new `DeviceFingerprintInput` union type to match runtime behavior.
- Add regression unit tests that assert two different client fingerprint strings remain distinct and are not equal to the SHA-256 of an empty structured fingerprint while also asserting structured-object hashing still works.

### Testing
- Added unit test coverage in `packages/api/src/utils/__tests__/deviceUtils.test.ts` that validates preserved client fingerprint strings and structured hashing, but running Jest in this environment failed because `jest` is not available here (test run blocked).
- Attempted `bun run build` which failed in this environment due to unrelated TypeScript config errors in the `contracts` build (the build system in this environment raised `tsconfig` deprecation/rootDir errors), so full CI-style verification was not possible here.
- Local static checks (`git diff --check`) and the repository patch applied cleanly and the new tests were committed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db0e148323a1e4fe63930be570)

### DIFF
--- a/packages/api/src/types/session.types.ts
+++ b/packages/api/src/types/session.types.ts
@@ -4,8 +4,8 @@
  * Centralized type definitions for session-related operations.
  */
 
-import { ISession } from '../models/Session';
-import { DeviceFingerprint } from '../utils/deviceUtils';
+import type { ISession } from '../models/Session';
+import type { DeviceFingerprintInput } from '../utils/deviceUtils';
 
 export interface SessionValidationResult {
   session: ISession;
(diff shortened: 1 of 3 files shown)


NateIsern:
✅ Landed on main via 40d6f26a (wave-2 batch 2: session/billing/privacy/seed). api 735 tests green.

PR #296, opened 2026-06-25
fix(api): enforce refresh-token sibling checks

NateIsern:
### Motivation

- Prevent a refresh-token reuse-detection bypass where an attacker can present an unrelated used/revoked token alongside a valid token in the same `Cookie` header and suppress revocation containment. 
- Preserve the intentional migration tolerance for browser-owned duplicate cookies while restoring the security invariant that a presented used/revoked token from a different family/session must trigger reuse-detection.

### Description

- Tighten `classifyRefreshCandidates` in `packages/api/src/services/refreshToken.service.ts` to collect the first valid candidate and all used/revoked candidates and require that any used/revoked candidate be from the same `family`+`session` as the valid one before suppressing reuse-detection. 
- If any used/revoked candidate is from a different family/session the classifier now returns a `used` classification so the caller revokes the replayed family and deactivates the session. 
- Update the `/auth/refresh` route comment in `packages/api/src/routes/auth.ts` to document the same-family/session invariant. 
- Adjust `packages/api/src/routes/__tests__/refreshToken.test.ts` to (a) preserve same-family migration rotation behavior and (b) add a test that proves a used token from a different family presented alongside a valid token triggers reuse-detection and revocation.

### Testing

- Ran `git diff --check` to validate formatting and local diffs, which succeeded. 
- Attempted to run the refresh-token unit tests with `bun run test --runInBand packages/api/src/routes/__tests__/refreshToken.test.ts`, but the environment lacks `jest` so the run failed (command not found). 
- Attempted `bun install --frozen-lockfile` to bootstrap dependencies for local test execution, but it failed due to the CI environment returning HTTP 403 from the npm registry, so full test execution could not be completed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db255c8323a3347745821e24f2)

### DIFF
--- a/packages/api/src/routes/__tests__/refreshToken.test.ts
+++ b/packages/api/src/routes/__tests__/refreshToken.test.ts
@@ -436,10 +436,15 @@ describe('POST /auth/refresh', () => {
     const validCurrent = 'valid-current-token';
     const usedRow = buildStoredToken(usedStale, {
       _id: 'rt-used',
-      family: 'fam-used',
+      family: 'fam-shared',
+      sessionId: 'sess-shared',
       usedAt: new Date(Date.now() - 1000),
     });
-    const validRow = buildStoredToken(validCurrent, { _id: 'rt-valid', family: 'fam-valid' });
+    const validRow = buildStoredToken(validCurrent, {
+      _id: 'rt-valid',
+      family: 'fam-shared',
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #243, opened 2026-06-25
fix(auth): require RP grants for silent SSO

NateIsern:
### Motivation
- Prevent silent SSO from issuing a full Oxy session to a globally-approved RP when the user has not previously granted that RP, closing a security bypass that allowed approved RPs to obtain tokens on first visit.
- Ensure the IdP only performs silent token issuance for returning RPs that the user has explicitly granted, preserving interactive consent for first-time RPs.

### Description
- Add `userHasGrantedClient(apiBaseUrl, userId, clientOrigin)` helper that checks the per-user FedCM grants via `GET /fedcm/grants/:userId` and normalises origins.
- Require an existing per-user grant before minting or handing off sessions in the silent flows by gating `GET /auth/silent`, the central `GET /sso` bounce, and the per‑apex `GET /sso/establish` with `userHasGrantedClient` checks.
- Change behaviour for first-time (no-grant) but globally-approved RPs to return the same no-session/null-session outcome (`oxy_sso=none` or posted `null` session) so RPs must perform an interactive consent/sign-in flow instead of receiving tokens silently.
- Add regression tests to `packages/auth/server/__tests__/fedcm.idp.test.ts` that cover: same-apex SSO without a user grant, cross-apex establish without a user grant, and first-party `/auth/silent` without a grant.

### Testing
- Ran `git diff --check` which passed with no whitespace/patch issues.
- Attempted `bun test server/__tests__/fedcm.idp.test.ts` in `packages/auth` to exercise the IdP silent/S SO tests, but the local test run failed due to missing workspace dependency (`Cannot find package 'react'`), so the new tests were added but not executed successfully in this environment.
- Commit containing the fix was created: `fix(auth): require RP grants for silent SSO` (changes include `packages/auth/server/index.ts` and updated `packages/auth/server/__tests__/fedcm.idp.test.ts`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bcb698ba08323a6d4ff289dcdf4c5)

---

<a href="https://app.devin.ai/review/oxyhq/oxyhqservices/pull/243" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #297, opened 2026-06-25
fix(fedcm): gate user grants lookup

NateIsern:
### Motivation
- The unauthenticated `GET /fedcm/grants/:userId` endpoint exposed per-user FedCM grant metadata (which RP origins a user had previously granted) and could be queried by anyone knowing a user's ObjectId.
- This per-user relationship is private activity metadata and must be restricted to internal IdP→API server-to-server calls only.

### Description
- Require the internal shared secret header `X-Oxy-Internal` (validated in constant time) before returning per-user FedCM grants in `packages/api/src/controllers/fedcm.controller.ts`, and fail closed with `404` when the secret is missing/invalid.
- Update the FedCM OpenAPI docs and swagger components to document an `internalSecretAuth` scheme and remove the previous `security: []` (public) declaration for the grants endpoint in `packages/api/src/routes/fedcm.ts` and `packages/api/src/config/swagger.ts`.
- Change the IdP worker to present the internal secret when requesting user grants (send `X-Oxy-Internal`) and to skip the lookup when the secret is not configured, in `packages/auth/server/index.ts`.
- Add/adjust regression tests to assert missing/invalid secret cases and to verify the IdP sends the `X-Oxy-Internal` header, in `packages/api/src/controllers/__tests__/fedcm.controller.test.ts` and `packages/auth/server/__tests__/fedcm.idp.test.ts`.

### Testing
- Ran a static verification that the OpenAPI block for `/fedcm/grants/{userId}` no longer declares `security: []` and documents `X-Oxy-Internal` via a small Python check (`python3`), which succeeded.
- Attempted to run package tests with `bun run test` / `bun test` and `jest` for the modified controller, but automated test runs were blocked in this environment because `node_modules` and test runner deps (e.g. `jest`, `react`) are not installed; therefore the updated tests could not be executed here.
- Confirmed unit-test-level changes were applied (new assertions and env setup) and committed as `fix(fedcm): gate user grants lookup`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc60a88323b4a5a71e51355347)

### DIFF
--- a/packages/api/src/config/swagger.ts
+++ b/packages/api/src/config/swagger.ts
@@ -27,6 +27,12 @@ const options: swaggerJsdoc.Options = {
           bearerFormat: 'JWT',
           description: 'JWT access token obtained from /auth/login or /auth/verify',
         },
+        internalSecretAuth: {
+          type: 'apiKey',
+          in: 'header',
+          name: 'X-Oxy-Internal',
+          description: 'Internal shared secret for IdP/API server-to-server calls.',
+        },
       },
       schemas: {
         Error: {
--- a/packages/api/src/controllers/__tests__/fedcm.controller.test.ts
+++ b/packages/api/src/controllers/__tests__/fedcm.controller.test.ts
@@ -86,9 +86,11 @@ function createReq(overrides: Partial<Request> = {}): Request {
 }
 
 const VALID_USER_ID = '64f7c2a1b8e9d3f4a1c2b3d4';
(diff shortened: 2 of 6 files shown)


NateIsern:
✅ Landed on main (reworked to coexist with #243/#253). auth IdP 69/0; api fedcm 35 pass; api tsc 0. IdP already carries SSO_INTERNAL_SECRET (for /sso/code) so no new deploy dependency.

PR #299, opened 2026-06-25
fix(services): scope payment queries by session

NateIsern:
### Motivation
- Payment and wallet React Query hooks used a stable singleton cache key (e.g. `['payments','wallet','current']`) and the query persistence whitelist included `payments`, allowing cached billing/wallet data to be reused across account/session switches and across restarts.\
- This could leak sensitive financial data to a different active account in the same browser/profile until the cache expired or a refetch occurred.\

### Description
- Scope all current-user payment-related query keys to the active session by passing `activeSessionId` into `queryKeys.payments.*` usage (`packages/services/src/ui/hooks/queries/usePaymentQueries.ts`).\
- Remove `payments` from the persistence whitelist so successful payment reads are not dehydrated and persisted across cold restarts (`packages/services/src/ui/hooks/queryClient.ts`).\
- Clear in-memory payments queries during a session switch so cached entries are removed before activating the new session (`packages/services/src/ui/hooks/useSessionManagement.ts`).\
- Update unit tests to assert session-scoped payment keys and that wallet queries refetch after an active session change (`packages/services/__tests__/hooks/queryKeys.test.ts` and `packages/services/__tests__/hooks/usePaymentQueries.test.tsx`).\

### Testing
- Ran `git diff --check` to validate there are no whitespace/merge artifacts and it passed.\
- Updated and added unit tests (`queryKeys.test.ts` and `usePaymentQueries.test.tsx`) that exercise the new session-scoped keys and a wallet refetch-on-session-switch behavior, but running them locally was blocked because the test runner (`jest`) is not available in this environment.\
- Attempted `bun install` to run the full package tests but dependency fetches failed (registry returned `403`), preventing end-to-end automated test execution in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db8f508323abcb8c61581c332f)

### DIFF
--- a/packages/services/__tests__/hooks/queryKeys.test.ts
+++ b/packages/services/__tests__/hooks/queryKeys.test.ts
@@ -90,17 +90,19 @@ describe('queryKeys.payments', () => {
     expect(queryKeys.payments.all).toEqual(['payments']);
   });
 
-  it("subscription() resolves to 'current' when no userId is provided", () => {
+  it("subscription() resolves to 'current' when no session scope is provided", () => {
     expect(queryKeys.payments.subscription()).toEqual(['payments', 'subscription', 'current']);
-    expect(queryKeys.payments.subscription('u1')).toEqual(['payments', 'subscription', 'u1']);
+    expect(queryKeys.payments.subscription('sess-1')).toEqual(['payments', 'subscription', 'sess-1']);
   });
 
-  it('history() builds [payments, history, current]', () => {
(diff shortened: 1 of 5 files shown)


NateIsern:
Closing as duplicate of #301 (payment query-cache scoping). #301 scopes by user id, keeps offline persistence, AND bumps the persisted-cache key version to purge the stale unscoped blob. #299 drops offline persistence. #301 is the winner.

PR #300, opened 2026-06-25
fix(services): scope payment query cache by account

NateIsern:
### Motivation
- Prevent cross-account leakage of sensitive payment and wallet data caused by payment queries using unscoped static keys (e.g. `['payments','history','current']`) that are shared across local account/session switches.  
- Ensure payment-related React Query entries are account-scoped so switching `activeSessionId` + `user.id` cannot surface cached data from a previously active account.

### Description
- Scope payment query keys by an account identifier derived from `activeSessionId` and `user.id` (named `accountScope`) for subscription, payment history, wallet, and wallet transactions so each account has its own cache entries; updated `useUserSubscription`, `useUserPayments`, `useUserWallet`, and `useUserWalletTransactions` accordingly in `packages/services/src/ui/hooks/queries/usePaymentQueries.ts`.
- Tighten hook enablement so queries remain disabled until a valid `accountScope` exists (`enabled: ... && !!accountScope`) to avoid rendering stale shared cache entries during session transitions.
- Include `accountScope` in the pagination-aware `walletTransactions` query key so pages remain per-account and per-page.
- Add a regression unit test `scopes payment cache entries by active user to prevent cross-account reuse` to `packages/services/__tests__/hooks/usePaymentQueries.test.tsx` that mocks two different accounts/sessions and verifies separate fetches and distinct cached results.

### Testing
- Added a unit test in `packages/services/__tests__/hooks/usePaymentQueries.test.tsx` that asserts a switched account triggers a second fetch and does not reuse the first account's cached payments.  
- Ran `git diff --check` with no whitespace errors.  
- Attempted to run the package tests (`jest`) in this environment, but the `jest` binary is not available here so automated test execution was blocked (test addition is present and designed to pass in CI where `jest` is installed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dbcc948323b69d8dd7f6846f8e)

### DIFF
--- a/packages/services/__tests__/hooks/usePaymentQueries.test.tsx
+++ b/packages/services/__tests__/hooks/usePaymentQueries.test.tsx
@@ -194,6 +194,37 @@ describe('payment query hooks', () => {
     });
   });
 
+  it('scopes payment cache entries by active user to prevent cross-account reuse', async () => {
+    mockState.oxyServices.getUserPayments
+      .mockResolvedValueOnce([{ ...PAYMENTS_FIXTURE[0], id: 'p-u1', userId: 'u1' }])
+      .mockResolvedValueOnce([
+        { ...PAYMENTS_FIXTURE[0], id: 'p-u2', userId: 'u2', amount: 19.99 },
+      ]);
+
+    const first = renderHook(() => useUserPayments(), {
+      wrapper: makeWrapper(queryClient),
+    });
+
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as duplicate of #301. #300's ${session}:${user.id} composite key is redundant (user id alone discriminates) and it drops payments from offline persistence. #301 is the cleaner winner.

PR #301, opened 2026-06-25
fix(services): scope payment query cache by user

NateIsern:
### Motivation
- Fix a client-side confidentiality leak where payment/wallet/subscription queries used shared `current` query keys and the `payments` namespace was persisted, allowing one account's financial data to be visible after session switching or hydration. 
- Ensure financial data is either not persisted across sessions or is isolated per authenticated principal.

### Description
- Scope payment-related query keys to the authenticated user id by passing `user?.id` into `queryKeys.payments.*` and require a resolved `user?.id` before enabling the hooks in `packages/services/src/ui/hooks/queries/usePaymentQueries.ts`.
- Remove the `'payments'` namespace from the persisted query whitelist and bump the persisted cache key to `oxy_query_cache_v3` while retaining legacy keys for explicit clears in `packages/services/src/ui/hooks/queryClient.ts`.
- Add regression tests that assert the hooks do not run without a scoped user id and that wallet cache entries are stored per-user in `packages/services/__tests__/hooks/usePaymentQueries.test.tsx`.
- Files changed: `usePaymentQueries.ts`, `queryClient.ts`, and `__tests__/hooks/usePaymentQueries.test.tsx` (plus small test/import adjustments).

### Testing
- Ran `git diff --check` which reported no whitespace or merge issues (success).
- Attempted to run the package tests with `bun run test -- usePaymentQueries queryKeys --runInBand`, but the runner failed in this environment because `jest` is not available (`jest: command not found`), so the new tests were added but not executed here.
- Attempted a TypeScript compile with `bun run typescript` which failed due to missing workspace dependencies/type declarations in this execution environment (e.g. `react`, `@oxyhq/core`, `react-native`, `@tanstack/react-query`), so local/CI compilation should be validated in an environment with workspace deps installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db5b988323b1a533b160639dfb)

### DIFF
--- a/packages/services/__tests__/hooks/usePaymentQueries.test.tsx
+++ b/packages/services/__tests__/hooks/usePaymentQueries.test.tsx
@@ -132,6 +132,7 @@ import {
   useUserWallet,
   useUserWalletTransactions,
 } from '../../src/ui/hooks/queries/usePaymentQueries';
+import { queryKeys } from '../../src/ui/hooks/queries/queryKeys';
 
 const makeWrapper = (queryClient: QueryClient) =>
   function Wrapper({ children }: { children: ReactNode }) {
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #302, opened 2026-06-25
fix(api): bound app-data namespace reads

NateIsern:
### Motivation
- The per-user app-data KV feature allowed unlimited keys per namespace and returned an unpaginated full-namespace listing, creating a resource-exhaustion (DoS) risk when an attacker accumulated many values.
- Introduce defensive quotas and bounded listing to prevent large DB reads, excessive memory use, and huge JSON responses while preserving existing per-value size and write-rate limits.

### Description
- Add new constants `APP_DATA_MAX_NAMESPACE_KEYS` and `APP_DATA_MAX_USER_KEYS` to `packages/api/src/schemas/userData.schemas.ts` to define quotas (128 keys per namespace, 1024 keys per user).
- Enforce quotas in the write path by adding `enforceAppDataKeyQuotas(userId, namespace, key)` which checks counts and throws `ConflictError` for new keys; existing keys may still be updated.
- Bound namespace listings with `.limit(APP_DATA_MAX_NAMESPACE_KEYS + 1)` and return a `413 PAYLOAD_TOO_LARGE` `ApiError` when a namespace exceeds the quota instead of materializing an unbounded response.
- Add regression tests in `packages/api/src/routes/__tests__/userData.test.ts` covering quota rejection on PUT and oversized namespace listing behavior.

### Testing
- `git diff --check` was run locally and reported clean diffs (no whitespace/errors).
- New Jest route tests were added (`packages/api/src/routes/__tests__/userData.test.ts`) to validate quota enforcement and bounded-list behavior, but they could not be executed in this environment because test dependencies are not installable here (`jest` not found).
- `bun install --frozen-lockfile` was attempted but failed due to upstream registry `403` errors, so automated test execution was blocked; the added tests are deterministic and mock the DB layer to run under normal CI where dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db6d1c832399e4bdc9671852c2)

### DIFF
--- a/packages/api/src/routes/__tests__/userData.test.ts
+++ b/packages/api/src/routes/__tests__/userData.test.ts
@@ -1,0 +1,129 @@
+import express from 'express';
+import http from 'http';
+import { AddressInfo } from 'net';
+
+const mockAuthMiddleware = jest.fn();
+const mockExists = jest.fn();
+const mockCountDocuments = jest.fn();
+const mockFindOneAndUpdate = jest.fn();
+const mockFind = jest.fn();
+
+jest.mock('../../middleware/auth', () => ({
+  authMiddleware: (...args: unknown[]) => mockAuthMiddleware(...args),
+}));
+
+jest.mock('../../middleware/rateLimiter', () => ({
+  rateLimit: () => (_req: unknown, _res: unknown, next: () => void) => next(),
+}));
+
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #303, opened 2026-06-25
fix(api,security): avoid pre-auth login lockout denial

NateIsern:
### Motivation
- Prevent a targeted account lockout DoS where an unauthenticated attacker could fill an identifier's failure bucket and cause even a correct password to be rejected before verification.
- Preserve constant-time username/password timing hardening (H5) and retain throttling for repeated invalid credentials while ensuring legitimate users can still authenticate to clear failures (H7).

### Description
- Removed the pre-credential `isLockedOut` check from `SessionController.signIn` so the route always performs user lookup and password verification before applying failure-based throttling (changed `packages/api/src/controllers/session.controller.ts`).
- Stopped importing `isLockedOut` from the lockout service and kept `recordFailure`/`clearFailures` so failed verifications still increment the bucket and successful sign-ins clear it (changed `packages/api/src/controllers/session.controller.ts`).
- Updated sign-in regression tests to assert that the pre-check is not used and that credential verification still runs (changed `packages/api/src/controllers/__tests__/signIn.controller.test.ts`).
- Preserved existing behavior where failed verifications return `401` and once the threshold is reached subsequent invalid attempts return `429` with a `Retry-After` header.

### Testing
- Ran `git diff --check` which reported no whitespace or diff issues.
- Attempted to run the focused unit tests via `bun run test --runInBand src/controllers/__tests__/signIn.controller.test.ts`, but the test runner (`jest`) was not available in this environment so the test invocation could not execute.
- Attempted `bun install` to install deps required for running tests, but dependency fetches failed against the npm registry (multiple `403` responses), preventing full automated test execution in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dbd3848323be76b834fe78ab67)

### DIFF
--- a/packages/api/src/controllers/__tests__/signIn.controller.test.ts
+++ b/packages/api/src/controllers/__tests__/signIn.controller.test.ts
@@ -5,10 +5,10 @@
  *  - H5: constant-time username lookup. When the identifier resolves to
  *    no user, we must still execute a password verification call so the
  *    response timing is indistinguishable from the wrong-password branch.
- *  - H7: per-account lockout. After repeated failures we return 429 +
- *    Retry-After, but keep the same generic message body so the
- *    unauthenticated caller cannot tell account-locked from
- *    invalid-credentials.
+ *  - H7: per-account invalid-attempt throttling. After repeated failures we
+ *    return 429 + Retry-After for invalid credentials, but a correct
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #304, opened 2026-06-25
fix(inbox): clear cached state before settings sign-out

NateIsern:
### Motivation
- Prevent leaked inbox data across local/multi-session account sign-outs when the settings Sign out path triggers a raw `logout()` without clearing inbox-specific caches and API instances.  
- Ensure the settings sign-out reuses the same inbox-reset semantics as the account-switcher flow to avoid showing previous-account messages, labels, or settings after an automatic session switch.

### Description
- Added a shared helper `resetInboxState()` in `packages/inbox/hooks/useEmail.ts` that clears the Zustand email UI state, nulls the cached `_api` instance, and calls `queryClient.clear()` to flush TanStack Query caches.  
- Imported and wired `resetInboxState()` into `packages/inbox/components/settings/sections/AccountSection.tsx` so the handler calls `resetInboxState()` before invoking `logout()`.  
- Kept existing behavior and UX of the settings sign-out dialog while ensuring inbox caches are cleared first.

### Testing
- Ran `git diff --check` to validate no whitespace errors; this completed successfully.  
- Ran a TypeScript check with `bunx tsc --noEmit -p packages/inbox/tsconfig.json`, which could not complete in this environment due to missing local dev dependencies and Expo TypeScript base; the typecheck was blocked by the environment rather than the change.  
- Ran the package lint script `bun run --cwd packages/inbox lint`, which could not run here because `expo` is not installed in the environment; linting should be executed in CI or a dev workstation and is expected to pass.  

If desired, run the following locally or in CI to fully validate: `bunx tsc --noEmit -p packages/inbox/tsconfig.json` and `bun run --cwd packages/inbox lint`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8daef18832386e372ede894673a)

### DIFF
--- a/packages/inbox/components/settings/sections/AccountSection.tsx
+++ b/packages/inbox/components/settings/sections/AccountSection.tsx
@@ -34,6 +34,7 @@ import { toast } from '@oxyhq/bloom';
 
 import { useColors } from '@/constants/theme';
 import { useSettings, useUpdateSettings } from '@/hooks/queries/useSettings';
+import { resetInboxState } from '@/hooks/useEmail';
 import type { EmailSettings } from '@/services/emailApi';
 import { SectionHeader } from '@/components/settings/SectionHeader';
 
@@ -151,6 +152,7 @@ export function AccountSection() {
 
   const handleSignOut = useCallback(async () => {
     try {
+      resetInboxState();
       await logout();
       toast.success('Signed out.');
     } catch (err) {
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as duplicate of #336 (inbox per-user cache isolation). #336's per-user query-key scoping + clear-on-identity-change already covers settings sign-out; both add a competing reset fn to useEmail.ts. #336 is the complete fix.

PR #305, opened 2026-06-25
fix(services): scope current user cache by session

NateIsern:
### Motivation

- Prevent a stale per-session cached `User` from being mirrored into the global auth store after account switches by ensuring current-user query data is keyed by `activeSessionId` when available.

### Description

- Make `queryKeys.accounts.current()` optionally accept a `sessionId` so current-user cache entries can be scoped to a session while preserving the unscoped singleton prefix when no session is provided.
- Switch `useCurrentUser` to use the session-scoped query key (`queryKeys.accounts.current(activeSessionId)`) so TanStack Query cache entries are separated per active session.
- Update all mutations, optimistic updates, rollbacks and helpers that read/write the current-user cache to use the session-scoped key (changes in `useAccountMutations.ts`, `mutationFactory.ts`, `useServicesMutations.ts`, and `avatarUtils.ts`).
- Add/adjust a unit test for `queryKeys.accounts.current()` to assert both the singleton prefix and the session-scoped variant (`packages/services/__tests__/hooks/queryKeys.test.ts`).

### Testing

- Ran `git diff --check` which passed with no whitespace errors.
- Attempted to run the focused test via `bun run test -- queryKeys.test.ts` but it failed because `jest` is not installed in this environment (command error: `jest: command not found`).
- Attempted to install dependencies with `bun install` but the job could not complete because npm registry tarball requests returned HTTP 403, preventing a full test run in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc339883239a5b7cec821027eb)

### DIFF
--- a/packages/services/__tests__/hooks/queryKeys.test.ts
+++ b/packages/services/__tests__/hooks/queryKeys.test.ts
@@ -29,8 +29,9 @@ describe('queryKeys.accounts', () => {
     expect(queryKeys.accounts.detail('s1')).toEqual(['accounts', 'detail', 's1']);
   });
 
-  it('current() is a stable singleton key', () => {
+  it('current() is a stable singleton prefix and scopes by sessionId when supplied', () => {
     expect(queryKeys.accounts.current()).toEqual(['accounts', 'current']);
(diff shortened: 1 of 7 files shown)


NateIsern:

PR #307, opened 2026-06-25
fix(api): constrain email import resource usage

NateIsern:
### Motivation
- The `/email/import` endpoint previously accepted up to 50 in-memory `.eml` uploads at 50 MiB each and parsed/uploaded attachments without calling `enforceQuota()`, enabling authenticated users to exhaust heap, CPU, and storage quotas and cause DoS. 

### Description
- Added import-specific multer limits and validation by introducing `IMPORT_MAX_FILES = 5`, `IMPORT_MAX_FILE_SIZE_BYTES = 25 * 1024 * 1024`, `fields: 0`, `parts` caps, and a `fileFilter` to reject non-`.eml` filenames early in `packages/api/src/routes/email.ts`. 
- Replaced the direct `upload.array(...)` usage with `importUploadMiddleware` that converts multer errors into the existing `BadRequestError` shape so early rejections are handled by the global error handler. 
- Enforced quota checks in `emailService.importMessages` by calling `enforceQuota()` on the total incoming buffers and again per-parsed message before uploading attachments, and validated each parsed attachment against the plan `maxAttachmentSize` from `EMAIL_QUOTAS` to prevent oversized attachments. 
- Preserved existing behavior for valid imports (parsing with `simpleParser`, `assetService.uploadFileDirect` uploads, `Message.create`, mailbox size increment, linking assets, and async AI processing) while failing early for invalid/over-quota imports. 

### Testing
- Ran `git diff --check`, which passed with no style/format errors. 
- Attempted `bun run build`, which failed before the API build due to the `@oxyhq/contracts` TypeScript build errors (TS5107/TS5011) in the current environment. 
- Attempted to run the package unit tests with `bun run test` / `jest`, but `bun install` failed due to npm registry 403 responses and `jest` was not available, so unit tests could not be executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc02088323baee453f359f5452)

### DIFF
--- a/packages/api/src/routes/email.ts
+++ b/packages/api/src/routes/email.ts
@@ -5,11 +5,12 @@
  * All routes require authentication via authMiddleware.
  */
 
-import { Router } from 'express';
+import { Router, type NextFunction, type Request, type Response } from 'express';
 import multer from 'multer';
 import { authMiddleware } from '../middleware/auth';
 import { asyncHandler } from '../utils/asyncHandler';
 import { validate } from '../middleware/validate';
+import { BadRequestError } from '../utils/error';
 import {
   createMailboxSchema,
   mailboxIdParams,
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #308, opened 2026-06-25
Protect inbound email webhook from unauthenticated rate-limit DoS

NateIsern:
### Motivation
- The public `/email/inbound` webhook applied a global rate limiter before authenticating requests and mounted a 25 MiB raw body parser ahead of authentication, allowing unauthenticated clients to exhaust the delivery quota and force large body buffering. 
- The Cloudflare Email Worker treated any 4xx response as a permanent rejection, which caused legitimate deliveries to bounce when the API returned `429` from an exhausted quota.

### Description
- Exported an authenticated-only verifier `verifyEmailInboundWebhookSecret` and the `inboundRateLimit` from `packages/api/src/routes/emailInbound.ts` and implemented a constant-time secret comparison for the bearer secret. 
- Moved the inbound-route middleware mount in `packages/api/src/server.ts` to `verifyEmailInboundWebhookSecret -> inboundRateLimit -> express.raw({ type: '*/*', limit: '25mb' })` so authentication and rate-limiting run before raw body buffering. 
- Removed the pre-auth route-level limiter from the handler so only authenticated, rate-limited requests hit the handler. 
- Updated `workers/email-inbound/src/index.ts` to treat `408` and `429` (timeouts and rate-limit) as retryable failures rather than permanent 4xx rejections so Cloudflare will retry deliveries instead of bouncing immediately.

### Testing
- Ran targeted source assertions via a Node script that validated the exported verifier/limiter, verified the middleware mount order is `auth -> rate-limit -> raw parser`, ensured the route-level limiter was removed, and confirmed the worker now excludes `429` from permanent rejections; these assertions passed. 
- Ran `git diff --check` to ensure no whitespace or obvious diff errors; this succeeded. 
- Attempted `bun run --cwd packages/api build` which is blocked by existing TypeScript config issues unrelated to this patch (contracts `tsconfig` deprecation/rootDir errors), so a full type-checked build could not be completed in this environment. 
- Attempted package tests (`jest`) but the environment lacks `jest` installation, so unit test execution was not possible here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db91088323ba478b7da31edb3a)

### DIFF
--- a/packages/api/src/routes/emailInbound.ts
+++ b/packages/api/src/routes/emailInbound.ts
@@ -14,7 +14,8 @@
  */
 
 import { Router } from 'express';
-import type { Request, Response } from 'express';
+import type { NextFunction, Request, Response } from 'express';
+import { timingSafeEqual } from 'node:crypto';
 import { simpleParser } from 'mailparser';
 import type { ParsedMail } from 'mailparser';
 import { rateLimit } from '../middleware/rateLimiter';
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #312, opened 2026-06-25
fix(api): sanitize federated profile text to prevent stored XSS

NateIsern:
### Motivation
- Prevent stored XSS introduced by decoding HTML entities from untrusted ActivityPub actor `name`/`summary` and then persisting the resulting raw markup into profile fields.
- Restore the repository invariant that profile text fields are HTML-escaped before storage so clients never receive unescaped markup from federation resolution.

### Description
- Import the existing HTML escaper `sanitizeHtml` and add `sanitizeFederatedProfileText` which decodes entities, strips revealed or literal tags when requested, and then re-escapes the result with `sanitizeHtml`.
- Replace direct `decodeHtmlEntities(...)` usages in the ActivityPub actor extraction so `displayName` and `bio` become `sanitizeFederatedProfileText(...)` (bios use `stripTags: true`).
- Federated profiles returned by `fetchActorProfile` and later persisted by the upsert paths now carry sanitized values, preserving the prior storage/return invariant.
- Add a regression unit test that mocks a malicious actor payload (encoded tags and entities) and asserts the service returns sanitized `displayName` and `bio`, and add test mocks for the federation key-pair model used during service import.

### Testing
- Ran `git diff --check` which passed with no whitespace/format errors.
- Added/updated `packages/api/src/services/__tests__/federation.service.test.ts` to cover the regression case but test execution was blocked because dependency installation failed; `bun install` could not download npm tarballs (registry `403`), so `jest` was not available and `bun run test` could not run the suite.
- The change is minimal and focused only on sanitization of federated profile text to avoid behavioral regressions outside federation resolution paths.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db326883239a93c9cc2e975aec)

### DIFF
--- a/packages/api/src/services/__tests__/federation.service.test.ts
+++ b/packages/api/src/services/__tests__/federation.service.test.ts
@@ -24,6 +24,8 @@ const mockCacheInvalidate = jest.fn();
 const mockAssetFileContentExists = jest.fn();
 const mockAssetUploadFileDirect = jest.fn();
 const mockAssetDeleteFile = jest.fn();
+const mockKeyPairFindOne = jest.fn();
+const mockKeyPairCreate = jest.fn();
 
 process.env.AWS_ACCESS_KEY_ID ||= 'test-access-key';
 process.env.AWS_SECRET_ACCESS_KEY ||= 'test-secret-key';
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as duplicate of #316 (federated profile stored XSS). #316 sanitizes at the sinks and additionally covers the PUT /users/resolve ingestion path #312 misses; landing both would double-escape. #316 is the winner.

PR #314, opened 2026-06-25
fix(api): protect platform stats stream

NateIsern:
### Motivation
- The unauthenticated `/platform-stats` REST and `/platform-stats/stream` SSE endpoints could disclose aggregate metrics and cause DB load amplification because each SSE connection started its own 2s refresh and `fetchStats` did not deduplicate in-flight refreshes.
- The goal is to restrict access to staff-only, prevent cache-stampede DB queries, and limit the number of concurrent long-lived streams to protect availability.

### Description
- Require staff-only access for both REST and SSE by applying `router.use(authMiddleware, requireStaff)` to the `platform-stats` router.  
- Add in-flight refresh deduplication via a shared `inFlightStatsRefresh: Promise<any> | null` and factor the heavy queries into `refreshStats(now)` so concurrent callers share the same refresh Promise.  
- Add a per-process concurrent stream cap with `MAX_ACTIVE_STATS_STREAMS` and `activeStatsStreams` to return `429` when the cap is exceeded.  
- Prevent overlapping sends per connection with `sendInProgress` and a `closed` flag, and ensure the active stream counter and timers are cleaned up on `req.on('close')`.

### Testing
- Attempted build with `bun run --cwd packages/api build`, but it failed due to upstream TypeScript configuration issues in `@oxyhq/contracts` (tsconfig deprecation/rootDir errors), so the package build didn't complete.  
- Attempted type-check with `bunx tsc -p packages/api/tsconfig.json --noEmit`, but this was blocked by the repo TypeScript deprecation warnings/error conditions.  
- Attempted to run `bunx --bun -p typescript@5.9.2 tsc -p packages/api/tsconfig.json --noEmit` to use a specific TypeScript version, but fetching that package failed with registry `403`, so automated type checks and builds could not be completed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db91788323839adbc49827563c)

### DIFF
--- a/packages/api/src/routes/platform-stats.ts
+++ b/packages/api/src/routes/platform-stats.ts
@@ -1,4 +1,6 @@
 import { Router, Request, Response } from 'express';
+import { authMiddleware } from '../middleware/auth';
+import { requireStaff } from '../middleware/requireStaff';
 import { User } from '../models/User';
 import Session from '../models/Session';
 import { Message } from '../models/Message';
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #315, opened 2026-06-25
fix: enforce frozen lockfile in Docker builds

NateIsern:
### Motivation
- Prevent Docker builds from mutating or re-resolving dependencies during image creation so production images strictly reflect the committed `bun.lock` and avoid supply-chain drift.
- The previous build flow rewrote the root workspace metadata and ran installs without lockfile enforcement, which can admit unreviewed package resolution during deploy-time rebuilds.

### Description
- Restore lockfile-based installs by copying `bun.lock` into both the builder and production stages and running `bun install --frozen-lockfile` (and `bun install --production --frozen-lockfile` for runtime deps).
- Stop rewriting the root `workspaces` value inside the Docker build; instead copy every workspace `package.json` manifest before install so the monorepo package graph aligns with the committed lockfile.
- Keep source & built artifact copy steps intact and only change the dependency-install steps and the files staged into the image prior to installation.

### Testing
- Verified the Dockerfile no longer contains `BUN_INSTALL_FROZEN_LOCKFILE=false` and that `bun.lock` is copied into both stages via a repo search (`rg -n "bun.lock|--frozen-lockfile"`).
- Ran `git diff --check` to ensure no whitespace/index issues in the Dockerfile change and inspected the resulting Dockerfile lines to confirm the new `--frozen-lockfile` flags were present. 
- Attempted `bun install --frozen-lockfile` locally in the workspace to validate the command path, but the run was blocked by upstream registry `403` responses in this environment so a full install/build within CI was not executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db2f088323afe7ac59c02e7363)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,23 +14,25 @@ RUN npm install -g bun
 
 WORKDIR /app
 
-# Copy workspace root and override workspaces to only include api + core + contracts.
-# `@oxyhq/api` depends on `@oxyhq/contracts` (workspace:*); core is retained for the
-# admin scripts that import packages/core/src/* at runtime.
-# Remove bun.lock since the workspace change invalidates it — bun will
-# resolve fresh dependencies (still deterministic from package.json versions).
-COPY package.json ./
-RUN node -e "const p=require('./package.json'); p.workspaces=['packages/core','packages/contracts','packages/api']; require('fs').writeFileSync('package.json', JSON.stringify(p, null, 2));"
-
-# Copy package.json files for dependency resolution
+# Copy the root lockfile and every workspace manifest before installing.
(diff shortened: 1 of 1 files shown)


NateIsern:
Closing as not-correct — re-introduces a change already reverted on main. This restores the FULL-workspace Docker `bun install --frozen-lockfile` (copying accounts/auth/console/services manifests), the exact approach reverted in commit 83df9b39 because the full-workspace frozen install pulls esbuild whose arm64/alpine postinstall hard-fails ('Expected 0.27.2 but got 0.25.12'). The API image intentionally uses the lean core+contracts+api subset install.

PR #319, opened 2026-06-25
fix(inbox): pin markdown-it to 12.3.2 to mitigate ReDoS in Alia chat

NateIsern:
### Motivation
- The Inbox app pulls `@alia.onl/sdk` which brings `react-native-markdown-display` that resolves `markdown-it@10.0.0`, a version with known ReDoS/uncontrolled-resource-consumption vulnerabilities that can crash or hang client apps when parsing attacker-controlled Markdown.
- The change aims to force the markdown renderer onto a patched `markdown-it` to eliminate the vulnerable parser while preserving existing runtime behavior of the Alia chat surface.

### Description
- Added a root `overrides` entry for `markdown-it` pinned to `12.3.2` in `package.json` so all workspace installs prefer the patched parser version.
- Updated `bun.lock` entries to resolve `markdown-it` to `12.3.2` and adjusted its immediate transitive entries (`argparse`, `entities`, `linkify-it`) to compatible versions.
- The change is intentionally minimal and only forces a safe downstream `markdown-it` resolution so `react-native-markdown-display` used by the Alia SDK will use the patched parser.

### Testing
- Verified the lockfile no longer contains `markdown-it@10.0.0` and now contains `markdown-it@12.3.2` using scripted assertions, and the checks succeeded.
- Ran `rg` searches for the vulnerable/resolved versions and `bun pm ls markdown-it` to confirm the override is visible, and those checks succeeded.
- Validated `package.json` JSON syntax with `python3 -m json.tool` and ran `git diff --check`, both succeeded.
- Attempts to run `bun install` / `bun install --frozen-lockfile` to fully validate install-time behavior were blocked by the environment returning `403 Forbidden` from the npm registry, so full install verification could not be completed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db41208323941da978edfbd6e7)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -3441,7 +3441,7 @@
 
     "map-obj": ["map-obj@4.3.0", "", {}, ""],
 
-    "markdown-it": ["markdown-it@10.0.0", "", { "dependencies": { "argparse": "^1.0.7", "entities": "~2.0.0", "linkify-it": "^2.0.0", "mdurl": "^1.0.1", "uc.micro": "^1.0.5" }, "bin": "bin/markdown-it.js" }, ""],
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #322, opened 2026-06-25
fix(api): bound BIMI DNS avatar lookup

NateIsern:
### Motivation
- The BIMI TXT lookup path used `dns.resolveTxt()` without an application-level timeout, which allowed attacker-controlled sender domains with slow/unresponsive nameservers to stall avatar enrichment and delay mailbox API responses.

### Description
- Add a `BIMI_DNS_TIMEOUT_MS` constant and a `withTimeout` helper, and apply it to the BIMI lookup so `dns.resolveTxt()` is bounded in `packages/api/src/services/senderAvatar.service.ts`.
- Preserve existing fallback behavior so a timed-out or missing BIMI record falls through to the Gravatar and favicon checks unchanged.
- Add a Jest regression test at `packages/api/src/services/__tests__/senderAvatar.service.test.ts` that simulates a never-resolving BIMI TXT promise and verifies `getAvatarPath()` completes, returns `null`, and caches the `source: 'none'` result.
- Keep batch concurrency logic as-is; the bounded BIMI lookup prevents a single slow DNS response from blocking a batch indefinitely.

### Testing
- A new unit test was added at `packages/api/src/services/__tests__/senderAvatar.service.test.ts` to validate the timeout behavior, but automated test execution could not be performed in this environment because the `jest` test runner is not available here; CI is expected to run `bun run --cwd packages/api test` which will execute the new test.
- Build/typecheck attempts in this environment were blocked by missing local dev dependencies and existing TypeScript config issues unrelated to this change, so no further automated test runs were executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc15648323ab10e457b35580e9)

### DIFF
--- a/packages/api/src/services/__tests__/senderAvatar.service.test.ts
+++ b/packages/api/src/services/__tests__/senderAvatar.service.test.ts
@@ -1,0 +1,55 @@
+const mockResolveTxt = jest.fn();
+const mockSenderAvatarFindOne = jest.fn();
+const mockSenderAvatarUpdateOne = jest.fn();
+const mockUserFindOne = jest.fn();
+
+jest.mock('dns/promises', () => ({
+  __esModule: true,
+  default: { resolveTxt: (...args: unknown[]) => mockResolveTxt(...args) },
+}));
+
+jest.mock('../../models/SenderAvatar', () => ({
+  SenderAvatar: {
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #325, opened 2026-06-25
fix(api): bind analytics access to authenticated user

NateIsern:
### Motivation
- Close an authorization gap that allowed authenticated users to read or modify another user's analytics by passing a target `userID` in query/body while the premium check validated only the requester.
- Ensure analytics operations are always scoped to the authenticated principal so privacySettings and premium gating are enforced correctly.

### Description
- Changed analytics controllers to derive the target `userID` from the authenticated request (`req.user._id`) and to return `401` when no authenticated user id is present, removing trust in client-supplied `userID` values in queries and bodies (`packages/api/src/controllers/analytics.controller.ts`).
- Added a small helper `getAuthenticatedAnalyticsUserId` and adjusted controller handlers to use it for all reads/updates/aggregations so DB queries use the authenticated id only.
- Added unit tests that assert attacker-supplied `userID` values are ignored for `getAnalytics`, `updateAnalytics`, `getContentViewers`, and `getFollowerDetails`, and that access is rejected when `req.user` is missing (`packages/api/src/controllers/__tests__/analytics.controller.test.ts`).
- Kept behavior and response shapes unchanged for valid authenticated requests while hardening ownership checks server-side.

### Testing
- Ran `git diff --check` which reported no whitespace/error issues and the diffs were applied cleanly.
- Added controller unit tests under `packages/api/src/controllers/__tests__/analytics.controller.test.ts`, but running the test suite in this environment was blocked because `jest` was not present and `bun install` failed due to npm registry 403 responses, so the new tests were not executed here.
- Static verification performed by repository inspection and targeted mocks confirms the controllers now build queries against the authenticated user's id rather than client-supplied ids.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc66688323bcc1e4f508ee7215)

### DIFF
--- a/packages/api/src/controllers/__tests__/analytics.controller.test.ts
+++ b/packages/api/src/controllers/__tests__/analytics.controller.test.ts
@@ -1,0 +1,119 @@
+const mockAnalyticsFind = jest.fn();
+const mockAnalyticsAggregate = jest.fn();
+const mockAnalyticsFindOneAndUpdate = jest.fn();
+const mockUserFindById = jest.fn();
+const mockUserAggregate = jest.fn();
+
+jest.mock('../../models/Analytics', () => ({
+  __esModule: true,
+  default: {
+    find: mockAnalyticsFind,
+    aggregate: mockAnalyticsAggregate,
+    findOneAndUpdate: mockAnalyticsFindOneAndUpdate,
+  },
+}));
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #326, opened 2026-06-25
Gate AI thread summarization behind explicit user action to prevent automatic data disclosure

NateIsern:
### Motivation
- Prevent automatic disclosure of private email content to an external AI service when opening long threads by removing the unconditional render of the summarization component. 
- Provide an explicit opt-in affordance and a short disclosure so users know their conversation will be sent to the external summarization API.

### Description
- Stop automatically mounting `ThreadSummary` for threads of 4+ messages and instead require the user to tap a new "Generate AI thread summary" affordance before the component is mounted. (modified: `packages/inbox/components/MessageDetail.tsx`).
- Add `threadSummaryRequested` state to track user consent and conditionally mount `ThreadSummary` only after the user requests it. (modified: `packages/inbox/components/MessageDetail.tsx`).
- Add inline disclosure copy that explains the conversation will be sent to Alia and new styles for the prompt UI. (modified: `packages/inbox/components/MessageDetail.tsx`).

### Testing
- Ran `git diff --check` which reported no whitespace or diff-check issues and returned success. 
- Ran `bun run lint` in the inbox package which failed because the environment lacks `expo` (error: `expo: command not found`), so linter could not be fully validated in this environment.
- No unit/integration tests were executed in this environment as part of this change; recommend running the package's normal test/lint pipeline (`bun run lint`, `bun run test`) in CI or a development environment that has `expo` available to fully validate behavior and styling.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db55148323b741ba2836d99112)

### DIFF
--- a/packages/inbox/components/MessageDetail.tsx
+++ b/packages/inbox/components/MessageDetail.tsx
@@ -144,6 +144,7 @@ function MessageDetailInner({ mode, messageId }: MessageDetailProps) {
   const [replyTargetId, setReplyTargetId] = useState<string | null>(null);
   const [expandedMessages, setExpandedMessages] = useState<Set<string>>(new Set([messageId]));
   const [messageMenuId, setMessageMenuId] = useState<string | null>(null);
+  const [threadSummaryRequested, setThreadSummaryRequested] = useState(false);
 
   const moreMenuControl = useDialogControl();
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #327, opened 2026-06-25
fix(inbox): require explicit smart reply generation

NateIsern:
### Motivation

- Prevent automatic client-side transmission of private email bodies to the external Alia AI service by removing on-mount smart-reply generation. 
- Require explicit user consent before sending any message content to the AI and add stronger body-level sensitive-content checks to avoid exfiltrating OTPs, SSNs, card numbers, or other PII. 

### Description

- Gate smart-reply generation behind an explicit UI action by adding a `Generate quick replies with AI` button and `hasRequestedReplies` state in `packages/inbox/components/SmartReplyChips.tsx`, and only call `refetch()` after the user triggers it. 
- Disable the React Query for smart replies by default (`enabled: false`) in `packages/inbox/hooks/queries/useSmartReplies.ts` so mounting the component does not start the AI request. 
- Add `getMessageBody()` to normalize and strip HTML, then include body-aware sensitive filtering (keyword list + regex checks for SSNs, card-like numbers, OTP/code patterns) and use the sanitized body when building the prompt in `useSmartReplies.ts`. 
- Keep existing UX for showing/loading chips and inserting selected replies; add a short privacy disclosure next to the generate button and a few styling helpers in `SmartReplyChips.tsx`. 

### Testing

- Ran repository checks via `git diff --check` which produced no whitespace/format errors. 
- Ran the package lint script `bun run lint` in this environment but it failed because the `expo` CLI and other runtime dev deps are not installed here, so full lint/tests could not be executed. 
- No automated unit tests were run in this sandbox due to missing node modules and dev tooling; changes are limited, at-call sites `SmartReplyChips` now requires an explicit action before any network call and `useSmartReplies` never issues requests on mount.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc0bf88323a7002650f99e2bfe)

### DIFF
--- a/packages/inbox/components/SmartReplyChips.tsx
+++ b/packages/inbox/components/SmartReplyChips.tsx
@@ -5,12 +5,13 @@
  * Tapping a chip inserts the text into the reply composer.
  */
 
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import {
   View,
   Text,
   StyleSheet,
   Platform,
+  TouchableOpacity,
 } from 'react-native';
 import { Loading } from '@oxyhq/bloom/loading';
 import { Chip } from '@oxyhq/bloom/chip';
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #329, opened 2026-06-25
fix(inbox): harden native HTML email rendering to avoid WebView JS abuse

NateIsern:
### Motivation
- The native inbox renderer became reachable after adding `react-native-webview`, which allowed attacker-controlled email HTML to run JavaScript inside a native `WebView` and abuse the `postMessage` channel or active content. 
- The existing message handler accepted unbounded `height` values and the native `WebView` used a wildcard `originWhitelist`, enabling UI/availability attacks from malicious emails.

### Description
- Add `sanitizeEmailHtml` to strip `<script>` tags, active-content elements (iframes/objects/embeds/etc.), inline event handlers, and `javascript:` URL attributes before proxying external resources. 
- Add `MIN_WEBVIEW_HEIGHT` and `MAX_WEBVIEW_HEIGHT` constants and implement `clampWebViewHeight` to bound page-reported heights; the `onMessage` handler now applies the clamped height. 
- Replace the native `WebView` `originWhitelist` `['*']` with `['about:blank']` for static email HTML to reduce allowed navigation origins. 
- Use a min-height constant for native fallback rendering and keep existing navigation handling that opens external links in the system browser.

### Testing
- Ran `git diff --check` which succeeded with no whitespace errors. 
- Attempted `bun run --cwd packages/inbox lint` which could not complete in this environment because the `expo` CLI is not installed, so lint could not be fully validated. 
- Attempted `tsc` via `bunx --bun tsc -p packages/inbox/tsconfig.json --noEmit` which could not complete in this environment due to missing workspace/Expo TypeScript base configs and dependencies, so type-checking could not be fully validated here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db46988323bd934c6b070ceed2)

### DIFF
--- a/packages/inbox/components/HtmlBody.tsx
+++ b/packages/inbox/components/HtmlBody.tsx
@@ -17,6 +17,31 @@ interface HtmlBodyProps {
   html: string;
 }
 
+const MIN_WEBVIEW_HEIGHT = 100;
+const MAX_WEBVIEW_HEIGHT = 10_000;
+
+function sanitizeEmailHtml(html: string): string {
+  return html
+    .replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '')
+    .replace(/<\/?(?:iframe|object|embed|applet|form|input|button|textarea|select|meta|link)\b[^>]*>/gi, '')
+    .replace(/\s+on[a-z]+\s*=\s*(?:"[^"]*"|'[^']*'|[^\s>]+)/gi, '')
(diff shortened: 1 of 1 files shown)


NateIsern:
Closing as duplicate of #335 (inbox HtmlBody hardening). #329 keeps javaScriptEnabled on native; #335 is the stronger shared-util version (sanitizeEmailHtml + JS disabled + originWhitelist) and preserves the already-landed #258 Android-safe link path. #335 wins.

PR #331, opened 2026-06-25
fix(api): sanitize Alia streaming errors

NateIsern:
### Motivation
- The Alia proxy route allowed clients to request streaming (`req.body.stream === true`) and on upstream non-2xx responses would embed `err.response.data` (which can be a Node `IncomingMessage` stream) directly into `res.json()`, causing `TypeError: Converting circular structure to JSON` and an unhandled rejection that can crash the process.

### Description
- Add a `toSafeErrorMessage` helper and `isReadableStream` predicate to detect stream or otherwise non-serializable upstream error payloads and replace them with a safe fallback string instead of attempting to JSON-serialize them.
- Preserve existing streaming behavior (`responseType: 'stream'` and `response.data.pipe(res)`) for valid streaming responses while sanitizing error branches.
- Replace the previous direct use of `err.response?.data` in the catch block with the sanitized `toSafeErrorMessage(err.response?.data)` value.
- Add a regression test `packages/api/src/routes/__tests__/alia.test.ts` that mocks an Axios rejection containing an `IncomingMessage` body and asserts the route returns a safe JSON error.

### Testing
- Added a Jest regression test `packages/api/src/routes/__tests__/alia.test.ts` that verifies a streaming upstream error with a stream body returns `{ error: 'ALIA_PROXY_ERROR', message: 'Failed to reach Alia API' }` (test file committed).
- Attempted to run the test via `bun run test -- alia.test.ts` but `jest` is not available in the current environment so the test run could not execute (command failed).
- Attempted `bun install` to provision test deps but it failed due to npm registry `403` responses, so automated tests could not be executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc006c8323a677f2f6c68dfe88)

### DIFF
--- a/packages/api/src/routes/__tests__/alia.test.ts
+++ b/packages/api/src/routes/__tests__/alia.test.ts
@@ -1,0 +1,104 @@
+import express from 'express';
+import http, { IncomingMessage } from 'http';
+import { AddressInfo } from 'net';
+import axios from 'axios';
+
+jest.mock('axios');
+
+jest.mock('../../middleware/auth', () => ({
+  authMiddleware: (_req: unknown, _res: unknown, next: () => void) => next(),
+}));
+
+const mockedAxios = axios as jest.Mocked<typeof axios>;
+
+interface RawResponse {
+  status: number;
+  body: string;
+}
+
+const request = async (app: express.Express, body: unknown): Promise<RawResponse> => {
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #333, opened 2026-06-25
Fix unsafe email structured search regex filters

NateIsern:
### Motivation
- The email search endpoint accepted `from`, `to`, and `subject` query values and forwarded them directly into MongoDB `$regex` operators, enabling attacker-supplied pathological regexes to consume DB CPU and degrade availability.
- The intent is to keep structured search functionality while preventing executable regex injection and bounding query execution cost.

### Description
- Validate query parameters as single string values in `packages/api/src/controllers/email.controller.ts` by adding `getOptionalQueryString` and using it for `q`, `mailbox`, `from`, `to`, `subject`, `dateAfter`, and `dateBefore`.
- Escape user-supplied structured filters and enforce a length limit by adding `escapeRegexLiteral`, `normalizeStructuredSearchFilter`, and `MAX_STRUCTURED_SEARCH_FILTER_LENGTH = 128` to `packages/api/src/services/email.service.ts`, and use the normalized values for `from`, `to`, and `subject` regexes.
- Bound database execution time by adding `EMAIL_SEARCH_MAX_TIME_MS = 5000` and applying `.maxTimeMS(...)` to both the `find` and `countDocuments` queries in `packages/api/src/services/email.service.ts`.
- Add focused Jest unit tests at `packages/api/src/services/__tests__/emailService.searchMessages.test.ts` that assert filters are escaped, overlong filters are rejected, and `maxTimeMS` is applied.

### Testing
- Created and committed unit tests in `packages/api/src/services/__tests__/emailService.searchMessages.test.ts` (tests added but not executed in this environment).
- Ran `git diff --check` and a small Node snippet to validate `escapeRegexLiteral` behavior, both succeeded and confirmed regex metacharacters are escaped.
- Attempted to run the new Jest test via `bun run test` and `bun install`, but `jest`/dependencies could not be installed in this environment due to registry `403` errors, so Jest tests were not executed locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db98988323b9c90990e1b085bc)

### DIFF
--- a/packages/api/src/controllers/email.controller.ts
+++ b/packages/api/src/controllers/email.controller.ts
@@ -21,6 +21,12 @@ import {
 import { logger } from '../utils/logger';
 import type { RecipientInput, AttachmentInput } from '../schemas/email.schemas';
 
+function getOptionalQueryString(value: unknown, name: string): string | undefined {
+  if (value === undefined) return undefined;
+  if (typeof value === 'string') return value;
+  throw new BadRequestError(`${name} must be a single string value`);
+}
+
 /**
  * Resolve user-supplied { fileId, contentId?, isInline? } references into the
  * canonical IAttachment shape persisted on Message. Each fileId MUST:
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #335, opened 2026-06-25
Harden email HTML rendering in Inbox (sanitize + secure WebView)

NateIsern:
### Motivation
- Untrusted inbound email HTML was being rendered raw in the native WebView, leaving a path for script-based exfiltration and UI/phishing attacks. 
- The change introduces a conservative sanitization step and safer WebView defaults to reduce attack surface while preserving rich HTML rendering for benign emails. 

### Description
- Added `sanitizeEmailHtml(html: string)` in `packages/inbox/utils/htmlTransform.ts` that strips active tags, event handlers, and disallowed URL schemes and normalizes dangerous attributes. 
- Apply sanitization before proxying external resources and embedding HTML by updating `wrapHtml(...)` to proxy `sanitizeEmailHtml(html)` instead of the raw HTML in `packages/inbox/components/HtmlBody.tsx`. 
- Hardened the native WebView renderer by removing the injected height script, disabling JavaScript/DOM storage, restricting `originWhitelist` to `about:blank`, blocking in-WebView navigations, disabling cache/mixed content and multiple windows, and adding a conservative `minHeight` fallback. 

### Testing
- Ran a sanitizer smoke test that confirmed `<script>`, inline event handlers, `javascript:`/`data:` URLs and `<iframe>` elements are removed while safe `https://` image URLs are preserved. 
- Verified the code patterns (`javaScriptEnabled={false}`, `originWhitelist={['about:blank']}`, and `sanitizeEmailHtml`) are present with `rg` and checked there are no diff-check issues via `git diff --check`. 
- Attempted local lint/type checks, but `expo`/workspace dependencies and tsconfig resolution are not available in this environment so full `bunx tsc --noEmit` and `expo lint` could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8daf908832393831d606a7969ec)

### DIFF
--- a/packages/inbox/components/HtmlBody.tsx
+++ b/packages/inbox/components/HtmlBody.tsx
@@ -1,6 +1,6 @@
 /**
  * Cross-platform HTML email body renderer.
- * Web: sandboxed iframe with auto-height. Native: react-native-webview with auto-height.
+ * Web: sandboxed iframe with auto-height. Native: locked-down react-native-webview.
  *
  * External images and fonts are routed through our proxy to:
  * - Bypass CORS/CORP restrictions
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #338, opened 2026-06-25
fix(api): reject rotated session access tokens

NateIsern:
### Motivation
- Restore intended token-rotation semantics so a previously issued access token that has been rotated out of the session record is not accepted for authentication.
- Prevent attackers who possess an old but unexpired JWT from retaining access after the legitimate client has refreshed or reauthenticated.

### Description
- Add a timing-safe comparison helper `timingSafeTokenEqual` that uses `crypto.timingSafeEqual` to compare the presented access token with the session's stored `accessToken` in `packages/api/src/services/session.service.ts`.
- Update `validateSession` to invalidate the session cache and return `null` when the presented token does not match the active session's stored token, rejecting rotated-out tokens.
- Add a regression unit test `should reject a rotated-out access token for an active session` in `packages/api/src/services/__tests__/session.service.test.ts` to cover this behavior.

### Testing
- Ran `git diff --check` to validate the change formatting and diff sanity which passed locally.
- Attempted to run the focused unit test with `bun run test -- session.service.test.ts`, but execution was blocked because test dependencies (Jest) were not available in the environment.
- Attempted `bun install` to provision test dependencies, but the run failed due to npm registry tarball `403` responses, so automated tests could not be executed end-to-end in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dbf7b08323b04632b0add43ac6)

### DIFF
--- a/packages/api/src/services/__tests__/session.service.test.ts
+++ b/packages/api/src/services/__tests__/session.service.test.ts
@@ -381,6 +381,16 @@ describe('Session Service', () => {
 
       expect(result).toBeNull();
     });
+    it('should reject a rotated-out access token for an active session', async () => {
+      const mockSession = createMockSession({ accessToken: 'new-access-token' });
+      mockFindOneResults.push(mockSession);
+
+      const result = await sessionService.validateSession('old-access-token');
+
+      expect(result).toBeNull();
+      expect(sessionCache.invalidate).toHaveBeenCalledWith('session-123');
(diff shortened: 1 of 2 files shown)


NateIsern:
✅ Landed on main via 40d6f26a (wave-2 batch 2: session/billing/privacy/seed). api 735 tests green.

PR #340, opened 2026-06-25
fix: restrict credentialed SDK fetches

NateIsern:
### Motivation

- A prior change made broad SDK fetch helpers use `credentials: 'include'` for all requests, which can cause browsers to send HttpOnly cookies (e.g. `oxy_session_id` scoped to `.oxy.so`) to caller-supplied absolute URLs and leak bearer-equivalent session identifiers.

### Description

- Make `HttpService` only send ambient credentials when the request target origin matches the configured API `baseURL`; otherwise use `credentials: 'omit'` for fetch and `xhr.withCredentials = false` for RN XHR uploads.
- Add `getCredentialsMode(url)` and `shouldSendCredentials(url)` helpers to centralize origin checks for credentialed requests.
- Update `fetchAssetContent` to omit credentials for arbitrary asset URLs and only include them when the asset URL is on the configured API origin.
- Add unit tests that assert credentials are included for requests to the configured API origin and omitted for caller-supplied absolute URLs on other origins.

### Testing

- Ran `git diff --check` to validate no trailing whitespace or lefover diff issues, which passed. 
- Added unit tests under `packages/core/src/__tests__/httpServiceCsrf.test.ts` and `packages/core/src/mixins/__tests__/assetCredentials.test.ts` but executing the test runner was blocked because `jest` is not installed in this environment (`jest: command not found`).
- Attempted `bun run typescript` to typecheck, but it failed due to missing repository dependencies/types in the environment (errors from unresolved packages). 
- Attempted `bun install` but the environment could not fetch registry tarballs (npm registry returned 403), so dependency installation and full test/compile runs are blocked in this CI/development environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db66988323b36af1be8a5686ab)

### DIFF
--- a/packages/core/src/HttpService.ts
+++ b/packages/core/src/HttpService.ts
@@ -496,13 +496,21 @@ export class HttpService {
         const useXhrForUpload = isFormData && isReactNative() && typeof XMLHttpRequest !== 'undefined';
 
         const response = useXhrForUpload
-          ? await this.uploadViaXHR(fullUrl, method, headers, bodyValue as FormData, controller.signal, timeout)
+          ? await this.uploadViaXHR(
+              fullUrl,
+              method,
+              headers,
+              bodyValue as FormData,
+              controller.signal,
+              timeout,
+              this.shouldSendCredentials(fullUrl),
+            )
           : await fetch(fullUrl, {
               method,
               headers,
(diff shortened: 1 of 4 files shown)


NateIsern:

PR #341, opened 2026-06-25
fix(auth): keep recovery token in HttpOnly cookie

NateIsern:
### Motivation
- The recovery verification flow previously returned a raw `recoveryToken` in JSON and the client stored it in `sessionStorage`, exposing a bearer-equivalent token to same-origin JavaScript; this change closes that token exposure for the recovery path.
- This PR implements the safer cookie-based pattern already used elsewhere (HttpOnly, Secure, SameSite) so recovery flows do not leak bearer tokens to JS.

### Description
- Set an HttpOnly, SameSite=Lax, `Path=/auth/recover` cookie named `oxy_recovery_token` when a recovery code is verified instead of returning the raw token in JSON by adding `setRecoveryCookie(...)` and related helpers in `packages/api/src/controllers/session.controller.ts`.
- Make the reset endpoint read the recovery token from the server-side cookie and clear it after successful use by adding `getRecoveryTokenFromRequest(...)` and `clearRecoveryCookie(...)` in `packages/api/src/controllers/session.controller.ts`.
- Remove the `recoveryToken` field from the reset request schema so the server-side cookie is the authoritative bearer for reset (`packages/api/src/schemas/auth.schemas.ts`).
- Remove client-side `sessionStorage` handling and stop sending `recoveryToken` in the reset JSON body from the auth UI (`packages/auth/components/recover-form.tsx`).

### Testing
- Ran `git diff --check` which passed with the committed changes. 
- Verified no remaining client-side sessionStorage usage or JSON recovery-token echoes with `rg -n "sessionStorage\.(setItem|getItem|removeItem).*recovery|recoveryStorageKey|body: JSON.stringify(\{ recoveryToken|return res.json(\{\s*recoveryToken"` across the repo.
- Attempted to run package tests but they were blocked by the environment: `bun run --filter @oxyhq/api test` failed due to missing `jest` and `bun run --filter auth test` failed due to missing `jsdom`/`react` dependencies, so full automated test run could not be completed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce7f9048323a8dfbfd9c6fee5a9)

### DIFF
--- a/packages/api/src/controllers/session.controller.ts
+++ b/packages/api/src/controllers/session.controller.ts
@@ -61,6 +61,33 @@ const RECOVERY_CODE_LENGTH = 10;
 const RECOVERY_CODE_TTL_MS = 10 * 60 * 1000;
 const RECOVERY_TOKEN_TTL_MS = 15 * 60 * 1000;
 const MAX_RECOVERY_ATTEMPTS = 3;
+const RECOVERY_COOKIE_NAME = 'oxy_recovery_token';
+const RECOVERY_COOKIE_PATH = '/auth/recover';
+
+function recoveryCookieOptions(maxAge: number) {
+  return {
+    httpOnly: true,
+    secure: process.env.NODE_ENV === 'production',
+    sameSite: 'lax' as const,
+    domain: process.env.REFRESH_COOKIE_DOMAIN || 'oxy.so',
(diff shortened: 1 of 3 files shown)


NateIsern:

PR #342, opened 2026-06-25
fix(api): avoid self-revoking stable FedCM sessions

NateIsern:
### Motivation
- Prevent a regression where issuing a refresh cookie for a reused FedCM/stable session deactivates the very session that was just refreshed, breaking repeat FedCM sign-in/silent-auth flows.

### Description
- Update `packages/api/src/services/refreshToken.service.ts` so `revokeFamily` is called with `undefined` for the `sessionId` when the stored row's `sessionId` equals the supplied `sessionId`, avoiding self-deactivation while still revoking the old token family.
- Add a regression test in `packages/api/src/services/__tests__/issueAndSetRefreshCookie.test.ts` that asserts same-user slot replacement that reuses the same `sessionId` revokes the prior family but does not call `deactivateSession`.
- Small clarifying comment added to explain the FedCM/stable-session special-case behavior.

### Testing
- Ran `git diff --check` to validate the working tree and formatting, which passed without errors.
- Attempted to run the package tests with `bun run test -- src/services/__tests__/issueAndSetRefreshCookie.test.ts --runInBand`, but the test runner invocation failed because `jest` is not available in the execution environment (command error: `jest: command not found`).
- Attempted `bun install --frozen-lockfile` to install deps and enable running Jest, but the install failed due to the environment returning HTTP 403 from the npm registry, so automated tests could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db6600832386613813c922d22d)

### DIFF
--- a/packages/api/src/services/__tests__/issueAndSetRefreshCookie.test.ts
+++ b/packages/api/src/services/__tests__/issueAndSetRefreshCookie.test.ts
@@ -225,6 +225,26 @@ describe('issueAndSetRefreshCookie — slot allocation', () => {
     expect(mockDeactivateSession).toHaveBeenCalledWith('sess-a');
   });
 
+  it('does not deactivate the current session when same-user slot replacement reuses the sessionId', async () => {
+    const existingRaw = 'SAME-SESSION-TOK';
+    stage(buildStored(existingRaw, {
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #343, opened 2026-06-25
fix(services): scope account query caches by session

NateIsern:
### Motivation
- Prevent a client-side confidentiality leak where account-sensitive TanStack Query results (devices, storage usage) were cached under static keys and could be shown after switching sessions/accounts.
- Avoid persisting stale device metadata across restarts by ensuring query keys are scoped to the active session and acting-as account.

### Description
- Extend query key helpers so `devices.list(...)` and `storage.usage(...)` accept an `accountScope` discriminator instead of using static keys.
- Add an `accountQueryScope(activeSessionId, actingAs)` helper and include it in `useUserDevices()` and `useAccountStorageUsage()` query keys so cached results are unique per-active-session/acting-as pair.
- Require an active session for devices and storage queries by tightening their `enabled` guards to also check `!!activeSessionId`.
- Small behavioral change: device list and storage usage caches are now segregated per session/acting-as, preventing cross-account reuse and reducing the risk of persisted stale sensitive data.

### Testing
- Ran `git diff --check` which reported no whitespace or diff errors.
- Attempted package tests with `bun run --cwd packages/services test -- --runTestsByPath src/ui/hooks/queries/useServicesQueries.ts` but the test runner could not be executed in this environment because `jest` and other dev dependencies are not installed here.
- Attempted TypeScript build with `bun run --cwd packages/services typescript` but it failed in the current container due to missing workspace dependencies and type declarations (e.g. `react`, `@oxyhq/core`, React Native typings), so full compile/test could not be validated here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dc46108323a041cf126587508a)

### DIFF
--- a/packages/services/src/ui/hooks/queries/queryKeys.ts
+++ b/packages/services/src/ui/hooks/queries/queryKeys.ts
@@ -46,7 +46,7 @@ export const queryKeys = {
   devices: {
     all: ['devices'] as const,
     lists: () => [...queryKeys.devices.all, 'list'] as const,
-    list: (userId?: string) => [...queryKeys.devices.lists(), userId] as const,
+    list: (accountScope?: string) => [...queryKeys.devices.lists(), accountScope] as const,
     details: () => [...queryKeys.devices.all, 'detail'] as const,
     detail: (deviceId: string) => [...queryKeys.devices.details(), deviceId] as const,
   },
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #344, opened 2026-06-25
fix(core): make identity backup refresh atomic

NateIsern:
### Motivation
- Prevent a stale-previous-account backup from remaining after a successful overwrite and later being restored to silently switch the device back to the prior account.
- Ensure the final backup refresh is part of the atomic success contract so callers never observe success while a cross-account backup remains.

### Description
- Make the final backup refresh in `_persistIdentityAtomic` a required step by snapshotting the existing backup, attempting the refresh, and rolling back both backup and primary on failure in `packages/core/src/crypto/keyManager.ts`.
- Add a `_rollbackBackup` helper to restore the backup slot to its prior state when a partial refresh fails.
- Update the SecureStore fault-injection test harness to support one-shot failures and add a regression test that verifies an overwrite rejects and rolls back if the final backup refresh fails in `packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts`.

### Testing
- Ran `git diff --check`, which passed with no whitespace or diff errors.
- Attempted to run the new atomicity test via `bun run test` / Jest, but `jest` was not available in the environment so unit tests could not be executed.
- Attempted `bun install` to enable running tests, but dependency installation failed due to npm registry tarball HTTP 403 responses in this environment, preventing local test execution.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8dd05388323a19c4b350c67bc87)

### DIFF
--- a/packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts
+++ b/packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts
@@ -21,7 +21,7 @@ import { setPlatformOS } from '../../utils/platform';
 // Fault-injectable in-memory secure store. `failPlan` lets a test make a
 // specific (op, key) pair throw to simulate a keychain that fails mid-write or
 // is transiently locked.
-const failPlan: { failKey?: string; failOp?: 'set' | 'get' } = {};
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #346, opened 2026-06-25
fix(services): redact avatar crop error logs

NateIsern:
### Motivation
- Prevent production leakage of on-device/local image URIs that can appear inside Error objects from image measurement/manipulation APIs by avoiding logging raw errors to the global `logger`.
- Maintain existing user-facing behavior (toasts / error UI) while removing PII-adjacent file paths from production logs and telemetry.

### Description
- Added a `SENSITIVE_URI_PATTERN` and `sanitizeImageErrorMessage` to redact common local URI/path patterns from error messages in the avatar crop flow in `packages/services/src/ui/screens/AvatarCropScreen.tsx`.
- Updated `normalizeCropError` to return sanitized `Error.message` content for Error/string inputs instead of propagating the raw message.
- Replaced the `Image.getSize` error handler that passed the raw failure object into `logger.error` with a generic `Error('Image measurement failed')` when logging, preserving the existing user-facing toast and `measureError` state.
- Kept crop failure logging but now log the normalized/sanitized error so production logs do not contain local file paths.

### Testing
- Ran `git diff --check` to validate no whitespace or trivial diff issues and it passed.
- Attempted `bun run --filter @oxyhq/services lint` but the lint step could not run in this environment because the `biome` tool / workspace dependencies are not installed (command failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db134483239b5a07df736bd55c)

### DIFF
--- a/packages/services/src/ui/screens/AvatarCropScreen.tsx
+++ b/packages/services/src/ui/screens/AvatarCropScreen.tsx
@@ -168,13 +168,22 @@ function isCanvasLoadFailure(err: unknown): boolean {
     );
 }
 
+const SENSITIVE_URI_PATTERN =
+    /\b(?:file|content|ph|assets-library):\/\/\S+|\/(?:data|var|private|storage|sdcard|tmp)\/\S+/gi;
+
+function sanitizeImageErrorMessage(message: string): string {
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #349, opened 2026-06-25
fix: enforce blocks before public media access

NateIsern:
### Motivation
- Prevent authenticated viewers who have been blocked from fetching another user's public media via the `/:id/stream` and `/:id/download` paths by ensuring social block checks run for authenticated requests before the public-file fast-path returns allowed.

### Description
- Change `checkMediaAccess` in `packages/api/src/services/mediaPrivacyService.ts` so the public-file shortcut only short-circuits for unauthenticated requests (`file.visibility === 'public' && !context && !viewerUserId`), and ensure authenticated viewers run the bidirectional block check before public media is allowed.
- Preserve existing behavior for unauthenticated public access and for synthetic/system owners that do not look like ObjectIds (the `isUserBlocked` ObjectId guard remains in place).
- Add regression tests `packages/api/src/services/__tests__/mediaPrivacyService.test.ts` that cover: blocked authenticated viewers being denied, anonymous public access skipping block queries, and synthetic owner (`__federation__`) behavior.

### Testing
- Added unit tests `mediaPrivacyService.test.ts` to cover the regressed scenarios; test file added to the repo and validated syntactically.
- Attempted to run `bun run test -- mediaPrivacyService.test.ts`, but `jest` was not available in the environment so the test run could not be executed (`jest: command not found`).
- Attempted `bun install` to install dependencies, but the environment returned HTTP 403 for registry tarballs so dependencies could not be installed and CI-style test runs could not complete.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cb8db19f48323a347022ac1e7ea80)

### DIFF
--- a/packages/api/src/services/__tests__/mediaPrivacyService.test.ts
+++ b/packages/api/src/services/__tests__/mediaPrivacyService.test.ts
@@ -1,0 +1,96 @@
+const mockBlockFindOne = jest.fn();
+
+jest.mock('../../models/Block', () => ({
+  __esModule: true,
+  default: {
+    findOne: (...args: unknown[]) => mockBlockFindOne(...args),
+  },
+}));
+
+jest.mock('../../models/User', () => ({
+  User: {
+    findById: jest.fn(),
+  },
+}));
+
+jest.mock('../../models/Restricted', () => ({
+  __esModule: true,
+  default: {},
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #352, opened 2026-06-25
fix(api): normalize profile color before premium check

NateIsern:
### Motivation
- The `oxy` premium color check compared the raw submitted value to the exact string `"oxy"` before the Mongoose schema trimmed and lowercased the value, allowing case/whitespace variants to bypass subscription authorization.

### Description
- Add a `normalizeProfileColor` helper and normalize `color` updates with `trim().toLowerCase()` before running the premium-only check so the authorization uses the same normalization as the `User` schema.
- Persist the normalized color value (`normalizedValue`) into the filtered update set so the service and model behavior remain consistent.
- Add a Jest unit test `packages/api/src/services/__tests__/user.service.color.test.ts` that asserts `oxy`, `OXY`, `OxY`, and ` oxy ` are blocked for non-premium users and that ordinary colors are normalized before saving.

### Testing
- Ran source assertions that verify `normalizeProfileColor` is present and the premium check compares against the normalized value, which passed (`python3` assertion script).
- Added unit tests for the new behavior (`user.service.color.test.ts`) but executing `jest` in this environment failed because test dependencies are not installed; the test file demonstrates the expected coverage.
- Attempted `bun install --frozen-lockfile` and `bun run test` but dependency fetch failed due to registry HTTP 403 responses, so end-to-end test execution could not be completed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce7ce48832380da75019d2173ce)

### DIFF
--- a/packages/api/src/services/__tests__/user.service.color.test.ts
+++ b/packages/api/src/services/__tests__/user.service.color.test.ts
@@ -1,0 +1,78 @@
+jest.mock('../../models/User', () => ({
+  __esModule: true,
+  default: {
+    findById: jest.fn(),
+    findOne: jest.fn(),
+  },
+}));
+
+jest.mock('../../models/Subscription', () => ({
+  __esModule: true,
+  default: {
+    findOne: jest.fn(),
+  },
+}));
+
+jest.mock('../../utils/userCache', () => ({
+  __esModule: true,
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #353, opened 2026-06-25
fix(console): harden static file server against malformed paths and directory traversal

NateIsern:
### Motivation
- Close two vulnerabilities in `packages/console/serve.js`: an unhandled `decodeURIComponent(url.pathname)` that throws `URIError` (remote DoS) and an unsafe `filePath.startsWith(DIST_DIR)` containment check that can be bypassed with encoded slashes.

### Description
- Add safe decoding and containment checks by wrapping `decodeURIComponent` in a `try/catch` to return `400 Bad Request` for malformed percent-encoding, canonicalizing `DIST_DIR` with `path.resolve`, resolving candidate paths with `path.resolve(DIST_DIR, '.' + pathname)`, and enforcing containment via a helper that uses `path.relative`/`path.isAbsolute` instead of `startsWith` while preserving the existing fallback-to-`index.html` behavior.

### Testing
- Manual static-server probe: a request with malformed percent-encoding returns `400 Bad Request`, an encoded traversal request returns `403 Forbidden`, and the root `GET /` still returns `200` and serves `index.html`; running `bun run test` in `packages/console` could not be executed here because `vitest` is not available in the environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce80c60832395ece173af0f2779)

### DIFF
--- a/packages/console/serve.js
+++ b/packages/console/serve.js
@@ -1,10 +1,10 @@
 import { createServer } from 'node:http';
 import { readFile } from 'node:fs/promises';
-import { join, extname } from 'node:path';
+import { join, extname, resolve, relative, isAbsolute } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url));
-const DIST_DIR = join(__dirname, 'dist');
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #356, opened 2026-06-25
fix(api): require CSRF when bearer requests carry cookies

NateIsern:
### Motivation
- A recent change added a CSRF exemption for bearer tokens by decoding the JWT payload without verifying the signature, allowing an attacker to craft an unsigned token and bypass CSRF for browser requests that still carry victim cookies.\n- The aim is to ensure the CSRF exemption is safe for true server-to-server calls while preserving protection for any request that may be using ambient cookies or other browser-attached credentials.\n- Also harden token comparison to avoid mismatched-length exceptions during timing-safe comparisons.

### Description
- Restrict bearer-token CSRF exemption in `packages/api/src/middleware/csrf.ts` so it only skips CSRF when the request is a bearer-only call with no ambient cookies (`hasAmbientCookies` guard).\n- Prevent `X-Native-App` header from enabling header-only CSRF mode when ambient cookies are present by requiring no ambient cookies for native-app bypass as well.\n- Harden the double-submit comparison by checking buffer lengths before `crypto.timingSafeEqual` to avoid errors on mismatched token sizes.\n- Add regression tests in `packages/api/src/middleware/__tests__/csrf.test.ts` to cover: bearer-only requests (no cookies), bearer requests that carry ambient cookies (should be rejected without CSRF), and `X-Native-App` header attempts when cookies are attached (should be rejected).

### Testing
- Ran `git diff --check` on the edits and confirmed no whitespace/patch errors.\n- Added unit tests in `packages/api/src/middleware/__tests__/csrf.test.ts` that exercise the new guards (tests included in the change).\n- Attempted to run `bun run test --runInBand src/middleware/__tests__/csrf.test.ts` but could not execute unit tests because `jest` was not available in the environment (`jest: command not found`).\n- `bun install` failed in this environment due to registry/package tarball `403` responses, preventing dependency installation and local test execution.\n- `bun run build` could not complete due to a TypeScript configuration issue surfaced in `@oxyhq/contracts` (deprecated `moduleResolution=node10` / missing explicit `rootDir`), so a full build/test cycle could not be run here; the new tests should pass when run in CI or a developer environment with dependencies available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8f3a883239a289434812ebf2e)

### DIFF
--- a/packages/api/src/middleware/__tests__/csrf.test.ts
+++ b/packages/api/src/middleware/__tests__/csrf.test.ts
@@ -53,18 +53,58 @@ describe('verifyCsrfToken', () => {
     mockWarn.mockClear();
   });
 
-  it('allows state-changing requests with an explicit bearer token and no CSRF header', () => {
+  it('allows bearer-only state-changing requests with no CSRF header', () => {
     const { res, next } = runVerify({
       headers: {
         authorization: 'Bearer user-session-token',
       },
-      cookies: {
-        csrf_token: 'cookie-token',
-      },
+      cookies: {},
     });
 
     expect(next).toHaveBeenCalledTimes(1);
     expect(res.status).not.toHaveBeenCalled();
+  });
+
+  it('rejects bearer requests that also carry ambient cookies without a CSRF header', () => {
+    const { res, next } = runVerify({
+      headers: {
+        authorization: 'Bearer forged-token',
+        cookie: 'csrf_token=cookie-token; session=victim-session',
+      },
+      cookies: {
+        csrf_token: 'cookie-token',
+        session: 'victim-session',
+      },
+    });
+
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as not-correct — would cause an outage and contradicts the auth contract. The API authenticates writes ONLY via Authorization: Bearer (no cookie auth surface), but csrfProtection plants a csrf_token cookie on every browser; forcing CSRF when 'bearer + ambient cookie' would 403 ALL RP web bearer writes (accounts/console/inbox). Per AGENTS.md, bearer writes do NOT carry app CSRF. (The timingSafeEqual length-guard micro-fix can be cherry-picked separately.)

PR #364, opened 2026-06-25
fix(auth): emit Bloom RGB tokens as rgb CSS

NateIsern:
### Motivation
- Bloom changed its color-token contract to emit RGB channel triplets; the auth app still wrapped every token with `hsl(...)`, producing invalid or incorrect CSS on first paint and causing theme visual regressions.

### Description
- Update `packages/auth/lib/bloom-css.ts` to emit `rgb(${value})` instead of `hsl(${value})` when rendering Bloom preset tokens, and add a regression test in `packages/auth/lib/__tests__/bloom-css.test.ts` that asserts `getBloomThemeCSS` includes `--primary: rgb(` and does not contain `--primary: hsl(`.

### Testing
- Ran `git diff --check` which passed for the change; added a unit test file exercising `getBloomThemeCSS` but the full `bun run test` in `packages/auth` could not complete because `bun install` failed in this environment due to registry `403` errors (missing `react`, `jsdom`, and other test deps), so the new test could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce838448323b2ce0b9ce3f94d47)

### DIFF
--- a/packages/auth/lib/__tests__/bloom-css.test.ts
+++ b/packages/auth/lib/__tests__/bloom-css.test.ts
@@ -31,6 +31,12 @@ describe("bloom-css", () => {
         expect(css).toContain("--primary:")
     })
 
+    test("getBloomThemeCSS emits Bloom RGB token values as rgb() colors", () => {
+        const css = getBloomThemeCSS("oxy")
+        expect(css).toContain("--primary: rgb(")
+        expect(css).not.toContain("--primary: hsl(")
+    })
+
     test("applyColorPreset injects a scoped <style> element", () => {
         applyColorPreset("blue", "chooser-hover")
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as not-correct. It changes hsl(${value}) → rgb(${value}), but the Bloom APP_COLOR_PRESETS token values are HSL triples (e.g. '277 66% 56%'). rgb(277 66% 56%) is invalid CSS (hue + mixed %) → var(--primary) resolves transparent → unstyled auth app. The existing hsl() wrapper is correct. (Documented Bloom web CSS-var runtime-vs-static landmine — the test passes statically but the runtime color breaks.)

PR #365, opened 2026-06-25
fix(sso): guard storage cleanup failures

NateIsern:
### Motivation
- A `SecurityError` thrown when reading `window.sessionStorage` could escape the intended best-effort cleanup and abort logout/local session cleanup, so the SSO bounce cleanup must tolerate storage access failures.

### Description
- Replace the unsafe `hasSessionStorage()` check with a safe `getSessionStorage()` accessor that returns `null` on thrown errors and use the returned `Storage` object for operations in `packages/services/src/ui/utils/activeAuthuser.ts`.
- Move `window.sessionStorage` access inside the existing `try/catch` in `packages/auth-sdk/src/WebOxyProvider.tsx` so storage getter errors are swallowed as intended by the helper.
- Add a regression test in `packages/services/__tests__/utils/storageHelpers.test.ts` that stubs a throwing `sessionStorage` getter and asserts `clearSsoBounceState()` does not throw.

### Testing
- Ran `git diff --check` to validate the working tree (clean).
- Added a targeted Jest regression test for `clearSsoBounceState()` but running `bun run test -- --runTestsByPath __tests__/utils/storageHelpers.test.ts` was blocked because `jest`/package dependencies are not installed in this environment. 
- Attempted `bun run typescript` for `packages/auth-sdk` to validate types but it was blocked due to missing workspace dependencies in this checkout.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce9831483238f1ca29d4408a9d3)

### DIFF
--- a/packages/auth-sdk/src/WebOxyProvider.tsx
+++ b/packages/auth-sdk/src/WebOxyProvider.tsx
@@ -203,14 +203,16 @@ function silentSignInKey(oxyServices: OxyServices): string {
  * No-ops off-web and on any storage failure (best-effort).
  */
 function clearSsoBounceStateWeb(): void {
-  if (!isWebBrowser() || typeof window.sessionStorage === 'undefined') return;
-  const origin = window.location.origin;
+  if (!isWebBrowser()) return;
   try {
-    window.sessionStorage.removeItem(ssoAttemptedKey(origin));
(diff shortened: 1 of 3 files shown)


NateIsern:
✅ Landed on main via 5894a1c6 (wave-2 auth-sdk cluster). auth-sdk 68 tests + auth IdP component tests green.

PR #366, opened 2026-06-25
fix(api): avoid no-op avatar refresh throttling

NateIsern:
### Motivation
- Non-forced `/users/resolve` calls were unconditionally scheduling background avatar refresh jobs even when the user already had a stored file-id avatar, which caused the in-process throttle (`_lastAvatarAttemptAt`) to be consumed by no-op work and could suppress a subsequent forced refresh within the throttle window.

### Description
- Change `PUT /users/resolve` to compute `hasExistingStoredAvatar` and only call `federationService.scheduleAvatarRefresh(...)` when `forceAvatarRefresh` is true or there is no existing stored file-id avatar (file: `packages/api/src/routes/users.ts`).
- Update unit tests in `packages/api/src/routes/__tests__/usersResolve.test.ts` to assert that non-forced resolves skip scheduling when a stored avatar exists and still schedule when no stored avatar exists.
- Preserve existing behavior for forced refreshes and initial remote-avatar scheduling so functionality and response contract remain unchanged.

### Testing
- Updated Jest unit tests in `packages/api/src/routes/__tests__/usersResolve.test.ts` to cover the skip-case and initial-download case; the tests were updated in this PR but could not be executed here because `jest` was not available in the environment (`jest: command not found`).
- Performed repository checks including `git diff --check` and committed changes as `fix(api): avoid no-op avatar refresh throttling`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce808748323855e210828ec6aee)

### DIFF
--- a/packages/api/src/routes/__tests__/usersResolve.test.ts
+++ b/packages/api/src/routes/__tests__/usersResolve.test.ts
@@ -400,7 +400,7 @@ describe('PUT /users/resolve (C4)', () => {
     expect(mockUserFindOneAndUpdate).toHaveBeenCalledTimes(1);
   });
 
-  it('no refresh flag: still schedules an avatar download with force=false (scheduler decides to skip)', async () => {
+  it('no refresh flag: skips scheduling when an existing stored avatar would make the worker a no-op', async () => {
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #367, opened 2026-06-25
fix(api): accept first-party FedCM issuers

NateIsern:
### Motivation
- The auth worker now mints FedCM ID tokens with an issuer derived from the request host (per-apex `https://auth.<rp-apex>`), but the API verifier only accepted a single configured issuer, causing valid multi-domain tokens to be rejected during `/fedcm/exchange`.
- This change restores interoperability for the new Clerk-style multi-domain FedCM flow while keeping the existing central issuer override for tests/local dev.

### Description
- Added `MULTIPART_TLDS` and a `registrableApex` helper to safely derive an audience apex while avoiding multi-part public-suffix pitfalls in `packages/api/src/services/fedcm.service.ts`.
- Implemented `isAllowedFedCMIssuer(issuer, audience)` which accepts either the configured central `FEDCM_ISSUER` or the dynamic `https://auth.<audience-apex>` first-party issuer and used it in the token issuer validation path.
- Replaced the strict equality issuer check with the new allowlist helper and improved logging to show expected values and the token `aud` on failure.
- Added two regression tests in `packages/api/src/services/__tests__/fedcm.service.test.ts` to cover: acceptance of a valid first-party `auth.<audience-apex>` issuer and rejection of an unrelated dynamic issuer.

### Testing
- Added automated tests `accepts a first-party auth.<audience-apex> issuer for multi-domain FedCM` and `rejects a dynamic issuer that is not the auth sibling of the token audience` in `packages/api/src/services/__tests__/fedcm.service.test.ts` (these tests were added but not executed in this environment).
- Attempted to run the targeted test (`bun run test -- fedcm.service.test.ts`) but `jest` was not available in the container, causing the test run to fail.
- Attempted `bun install` to fetch test dependencies but the npm registry returned `403` errors for required packages (including `jest`/`ts-jest`), preventing full automated test execution in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce97a4083239e7f734e7faa023e)

### DIFF
--- a/packages/api/src/services/__tests__/fedcm.service.test.ts
+++ b/packages/api/src/services/__tests__/fedcm.service.test.ts
@@ -283,6 +283,27 @@ describe('FedCM exchangeIdToken (H9)', () => {
     });
   });
 
+  it('accepts a first-party auth.<audience-apex> issuer for multi-domain FedCM', async () => {
+    const token = mintToken({ iss: 'https://auth.party.example' });
+    const result = await fedcmService.exchangeIdToken(token, createReq(APPROVED_ORIGIN));
+
+    expect('error' in result).toBe(false);
+    if ('error' in result) return;
+    expect(result.sessionId).toBe('sess-123');
+    expect(mockCreateSession).toHaveBeenCalledWith('user-123', expect.anything(), {
+      deviceName: 'FedCM Sign-In',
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #371, opened 2026-06-25
fix(inbox): restore privacy settings route

NateIsern:
### Motivation
- A redesign removed the dynamic `/settings/[section]` catch-all route while the settings catalog still advertises a `/settings/privacy` path, leaving the Privacy UI unreachable and causing the catalog links to hit the not-found screen. 
- Restore the missing route and the Privacy UI component so the catalog-driven navigation maps to an actual subscreen and preserves the same authenticated/unauthenticated gate pattern as other settings screens.

### Description
- Added the page route `packages/inbox/app/(drawer)/(tabs)/settings/privacy.tsx` that implements the `/settings/privacy` typed-route and uses the same `SettingsScreenShell` + `SettingsAuthGate` pattern as other settings pages. 
- Restored the Privacy UI as `packages/inbox/components/settings/sections/PrivacySection.tsx` implementing tracking-protection and sender-trust rows, informational admonition, and disabled toggles until server-backed prefs exist. 
- The new route looks up its catalog entry via `SETTINGS_SECTIONS` so the catalog-driven `router.push()`/`router.replace()` paths remain consistent with the filesystem routes.

### Testing
- Ran the static mapping script that compares `SETTINGS_SECTIONS` paths to filesystem routes (`python` check), and it verified all catalog paths are present (success). 
- Ran a repository diff/sanity check (`git diff --check`) and the check passed (no whitespace/conflict markers). 
- Attempted an app build with `bun run inbox:build`, but the build was blocked by a registry access error fetching `expo` in this environment (`GET https://registry.npmjs.org/expo - 403`), so a full runtime build/export could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8d14483238f541601cefae1e4)

### DIFF
--- a/packages/inbox/app/(drawer)/(tabs)/settings/privacy.tsx
+++ b/packages/inbox/app/(drawer)/(tabs)/settings/privacy.tsx
@@ -1,0 +1,42 @@
+/**
+ * /settings/privacy — tracking protection and sender trust defaults.
+ * Requires authentication; otherwise renders the standard `SettingsAuthGate`.
+ */
+
+import React from 'react';
+import Head from 'expo-router/head';
+import { useOxy } from '@oxyhq/services';
+
+import { SettingsAuthGate } from '@/components/settings/SettingsAuthGate';
+import { SettingsScreenShell } from '@/components/settings/SettingsScreenShell';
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #372, opened 2026-06-25
fix(services): declare optional expo-haptics peer

NateIsern:
### Motivation
- A recent change added `await import('expo-haptics')` as a dynamic import in `AvatarCropScreen`, but `@oxyhq/services` did not declare `expo-haptics` anywhere, causing TypeScript/Metro module-resolution errors for consumers that do not install the package.
- The runtime `try/catch` is not sufficient because TypeScript and bundlers require the module to be resolvable at compile/bundle time, so an ambient declaration or package metadata is required so consumers can build without adding the peer unless they opt in.

### Description
- Added `expo-haptics` to `packages/services/package.json` as a `devDependency` and as a `peerDependency` (with an `optional` entry in `peerDependenciesMeta`) so the dependency is documented and treated as optional for consumers.
- Added an ambient TypeScript declaration file at `packages/services/src/types/expo-haptics.d.ts` that declares the minimal `expo-haptics` API used by the dynamic wrappers so the package typechecks when the optional peer is not installed.
- Updated the workspace lockfile metadata (`bun.lock`) entries for the `@oxyhq/services` workspace to reflect the new optional peer/dev dependency so workspace metadata remains consistent.

### Testing
- Ran `git diff --cached --check` and `python -m json.tool packages/services/package.json` which succeeded and validated JSON and whitespace sanity.
- Ran the package TypeScript check `bun run --filter @oxyhq/services typescript` in this environment and confirmed that the `expo-haptics` resolution errors were removed, though full typechecking still fails in this environment due to many unresolved workspace peers (expected in CI or a consumer environment with peers installed).
- Attempted `bun install --lockfile-only` / dependency resolution but the environment returned registry `403 Forbidden` errors so lockfile regeneration could not be exercised here; lockfile edits were limited to updating the existing workspace metadata to include `expo-haptics` as optional.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbceac7588323b003dd429b92f227)


NateIsern:

PR #374, opened 2026-06-25
fix(services): add fallback Bloom theme provider

NateIsern:
### Motivation

- Migrated overlays (`BottomSheetRouter`, `SignInModal`) now call Bloom's `useTheme()` but `OxyProvider` does not always mount a `BloomThemeProvider`, which can throw during render for consumers that only use `<OxyProvider>`.
- The repository already provides a compatibility hook that falls back when Bloom context is absent, but several migrated components bypassed that fallback and caused a client-side crash/regression.

### Description

- Add a local `BloomThemeFallback` to `packages/services/src/ui/components/OxyProvider.tsx` that calls Bloom's `useTheme()` and only installs a default `<BloomThemeProvider mode="system">` when no upstream provider exists.
- Import `BloomThemeProvider` and `useTheme as useBloomTheme` from `@oxyhq/bloom/theme` and wrap the overlay subtree (lazy overlays, `ToastOutlet`, etc.) in the fallback so overlays always have Bloom theme context when mounted by `OxyProvider`.
- Preserve existing behavior when an app already provides its own `BloomThemeProvider` by detecting and reusing the upstream provider rather than replacing consumer configuration.

### Testing

- Ran `git diff --check` to ensure no whitespace/merge errors and it passed.
- Attempted `bun run --cwd packages/services typescript` to typecheck the package but it was blocked by missing local dependencies/type declarations in the execution environment (e.g. `react`, `react-native`, `@oxyhq/bloom`).
- Attempted `biome`/lint checks (`bunx --bun @biomejs/biome`) but the environment failed to resolve the registry (HTTP 403), so automated lint validation could not complete here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce7fcb083239c6f822707dd54ae)

### DIFF
--- a/packages/services/src/ui/components/OxyProvider.tsx
+++ b/packages/services/src/ui/components/OxyProvider.tsx
@@ -4,6 +4,7 @@ import type { OxyProviderProps } from '../types/navigation';
 import { OxyContextProvider, type OxyContextProviderProps } from '../context/OxyContext';
 import { QueryClientProvider, focusManager, onlineManager } from '@tanstack/react-query';
 import { BloomDialogProvider } from '@oxyhq/bloom';
+import { BloomThemeProvider, useTheme as useBloomTheme } from '@oxyhq/bloom/theme';
 import { ToastOutlet } from '@oxyhq/bloom/toast';
 import { logger as loggerUtil } from '@oxyhq/core';
 import { setupFonts } from './FontLoader';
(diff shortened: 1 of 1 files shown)


NateIsern:
Closing as not-correct (per maintainer decision). Adding a try/catch BloomThemeProvider fallback inside OxyProvider reverses the documented architecture (OxyProvider deliberately does NOT wrap BloomThemeProvider; apps mount it themselves and hoist it above any splash/loading branch). The hook-in-try/catch render is also fragile. If a missing-provider crash is hit in practice, the fix is to hoist BloomThemeProvider in the app, not to swallow it in the SDK.

PR #375, opened 2026-06-25
fix(services): preserve timed-out sessions during restore

NateIsern:
### Motivation
- A parallelized stored-session restore used an 8s `Promise.race` timeout that treated timeout winners as invalid and persisted only completed validations, which could permanently drop valid secondary sessions when the API was slow.
- The intent is to avoid logging out valid sessions due to transient latency while still pruning explicitly invalid sessions to prevent 401 loops on restart.

### Description
- Treat per-session validation timeouts as a soft/unvalidated sentinel instead of `null`, by introducing `VALIDATION_TIMEOUT_RESULT` and `StoredSessionValidationResult` in `packages/services/src/ui/context/OxyContext.tsx`.
- Collect `unvalidatedSessionIds` for sessions that timed out and keep them in the persisted ID list instead of removing them immediately.
- Persist validated `ClientSession` objects while passing `preserveSessionIds` to `updateSessions` so timed-out IDs are preserved; extend `updateSessions` in `packages/services/src/ui/hooks/useSessionManagement.ts` with an optional `preserveSessionIds?: string[]` parameter that is included when writing stored session IDs.
- Ensure the active-session invalidation path also preserves soft-timed-out IDs when rewriting the stored session list.

### Testing
- Ran `git diff --check`, which passed without whitespace/errors.
- Attempted `bun run --filter @oxyhq/services test` but it was blocked by the environment (`jest` not available), so test run was not completed.
- Attempted `bun run --filter @oxyhq/services typescript` which reported many missing dependency/type errors (environment lacked `react`, `react-native`, `@oxyhq/core`, etc.), so TypeScript build could not be validated here.
- Attempted `bun run --filter @oxyhq/services lint` and `bun install`, both of which were blocked (missing `biome` and npm registry 403 respectively), so full CI-style validation could not be performed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8b0d88323a0e6e873f0c12771)

### DIFF
--- a/packages/services/src/ui/context/OxyContext.tsx
+++ b/packages/services/src/ui/context/OxyContext.tsx
@@ -265,13 +265,17 @@ const COOKIE_RESTORE_TIMEOUT = 3000;
 const COLD_BOOT_OVERALL_DEADLINE = 20000;
 
 /**
- * Per-session timeout (ms) for the parallel stored-session validation in
+ * Per-session soft timeout (ms) for the parallel stored-session validation in
  * `restoreStoredSession`. Each `validateSession` call races against this timer
  * so a single slow/offline session never blocks the whole startup validation
- * sweep — sessions that don't answer in time resolve to `null` (treated as
- * unvalidated) and the remaining sessions still settle.
+ * sweep. Sessions that don't answer in time are treated as unvalidated and kept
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #378, opened 2026-06-25
fix: preserve inbox message cache bodies on flag updates

NateIsern:
### Motivation
- Flag-update endpoints (`updateFlags`) return a partial message that omits detail-only fields (`text`, `html`, `headers`) and writing that response directly into caches can drop message bodies and leave filtered views (e.g., Starred) stale.

### Description
- Add `mergeMessageUpdate(old, updated)` to merge a flag-update response into an existing cached `Message` while preserving `text`, `html`, and `headers` and merging `flags`.
- Update `useToggleStar` to apply `mergeMessageUpdate` when syncing the server response into `['message', messageId]`, `['messages']` and `['thread']` caches instead of replacing cached messages with the partial response.
- Apply the same safe-merge behavior to `useToggleRead` so read-flag updates do not remove cached body/header fields.
- Reintroduce invalidation of `['messages']` queries in `useToggleStar` `onSettled` so filtered caches (for example the Starred view) reconcile with the server after optimistic updates.

### Testing
- Ran `git diff --check`, which passed with no whitespace errors.
- Attempted type-check with `bunx tsc --noEmit -p packages/inbox/tsconfig.json`, but it could not be completed in this environment due to missing Expo/dependencies so it was not run to completion.
- Attempted lint with `bun run --filter inbox lint`, but it could not be completed here because the `expo` CLI is not available in the execution environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8c1d08323955c07ce48e85e33)

### DIFF
--- a/packages/inbox/hooks/mutations/useMessageMutations.ts
+++ b/packages/inbox/hooks/mutations/useMessageMutations.ts
@@ -41,6 +41,21 @@ function removeMessageFromPages(
   };
 }
 
+/** Merge a server flag-update response into an existing cached message without dropping
+ * detail-only fields (body, headers) that the update endpoint does not return.
+ */
+function mergeMessageUpdate(old: Message | null | undefined, updated: Message): Message {
+  if (!old) return updated;
+  return {
+    ...old,
+    ...updated,
+    text: updated.text ?? old.text,
+    html: updated.html ?? old.html,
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #386, opened 2026-06-25
fix: align iOS keychain group with native docs

NateIsern:
### Motivation

- The repository documentation was changed to instruct iOS entitlements use `group.so.oxy.shared` while the SDK still hard-coded `group.com.oxy.shared`, causing shared-keychain writes to fail with entitlement mismatch on iOS and breaking native cross-app SSO.

### Description

- Update `IOS_KEYCHAIN_GROUP` in `packages/core/src/crypto/keyManager.ts` to `'group.so.oxy.shared'` and adjust the explanatory comment to match the documented format.
- Extend the test SecureStore mock in `packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts` to record `setItemAsync` calls and their options.
- Add a regression test in `packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts` that asserts shared identity writes pass `keychainAccessGroup: 'group.so.oxy.shared'` to SecureStore.

### Testing

- Verified repository references with `rg -n "group\.com\.oxy\.shared|group\.so\.oxy\.shared|IOS_KEYCHAIN_GROUP"` and confirmed the constant and tests now reference `group.so.oxy.shared` (search succeeded).
- Ran `git diff --check` to ensure no whitespace or diff issues (succeeded).
- Attempted to run the package test with `bun run --cwd packages/core test -- keyManager.atomicity.test.ts`, but the container environment lacks `jest`/dependencies so the test run could not be executed here; CI/package-local test runs should run this test successfully when dependencies are installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce90efc83238076006b9adf1df4)

### DIFF
--- a/packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts
+++ b/packages/core/src/crypto/__tests__/keyManager.atomicity.test.ts
@@ -22,6 +22,11 @@ import { setPlatformOS } from '../../utils/platform';
 // specific (op, key) pair throw to simulate a keychain that fails mid-write or
 // is transiently locked.
 const failPlan: { failKey?: string; failOp?: 'set' | 'get' } = {};
+const setCalls: Array<{
+  key: string;
+  value: string;
+  options?: Record<string, unknown>;
+}> = [];
 
 jest.mock(
   'expo-secure-store',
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #388, opened 2026-06-25
Fix: persist phone/address/birthday from personal-info editor

NateIsern:
### Motivation
- The personal-info editor exposed `phone`, `address`, and `birthday` fields but the client-side save path and server allowlist did not include those keys, causing silent no-ops when users tried to save them. The change wires the UI → SDK → API path so these fields persist.

### Description
- Wire client save/update flow: `useProfileEditing.saveProfile` and `updateField` now include `phone`, `address`, and `birthday` when building the `UserProfileUpdate` payload (`packages/services/src/ui/hooks/useProfileEditing.ts`).
- API model and persistence: add `address` and `birthday` fields to the `IUser` model and schema; `phone` already exists and is now selected for authenticated reads (`packages/api/src/models/User.ts`).
- Server allowlist and response shaping: add `phone`, `address`, and `birthday` to the profile update allowlist and include them in `formatUserResponse` only when returning authenticated `/users/me` responses via an `includePrivateFields` option (`packages/api/src/services/user.service.ts`, `packages/api/src/routes/users.ts`).
- Contracts and types: expose the new fields in the shared contract `UserProfileUpdate` / `UserResponse` and the core `User` interface so client + server agree on shapes (`packages/contracts/src/userResponse.ts`, `packages/core/src/models/interfaces.ts`).

### Testing
- `git diff --check` passed locally.
- `bun run services:build` was attempted but blocked by an external npm registry 403 for `react-native-builder-bob` so the services build could not complete.
- `bun run core:build` and `bun run --filter @oxyhq/api build` were attempted but blocked by existing TypeScript configuration deprecation/rootDir issues in the environment, so the TypeScript builds could not complete.
- Package tests (`bun run --filter @oxyhq/api test`, `bun run --filter @oxyhq/contracts test`) could not be run because `jest` is not available in the execution environment, so automated unit tests were not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8a3a083238f088d941636849b)

### DIFF
--- a/packages/api/src/models/User.ts
+++ b/packages/api/src/models/User.ts
@@ -162,6 +162,8 @@ export interface IUser extends Document {
   };
   bio?: string;
   description?: string;
+  address?: string;
+  birthday?: string;
   locations?: Array<{
     id: string;
     name: string;
@@ -448,6 +450,8 @@ const UserSchema: Schema = new Schema(
     },
     bio: { type: String },
     description: { type: String },
+    address: { type: String, trim: true },
+    birthday: { type: String, trim: true },
     locations: [{
       id: { type: String, required: true },
       name: { type: String, required: true },
--- a/packages/api/src/routes/users.ts
+++ b/packages/api/src/routes/users.ts
@@ -230,7 +230,10 @@ router.get(
     }
 
     logger.debug('GET /users/me', { userId: req.user.id });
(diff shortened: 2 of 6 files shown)


NateIsern:

PR #389, opened 2026-06-25
fix(services): include link metadata descriptions

NateIsern:
### Motivation
- A refactor replaced the old Edit Links modal with `EditProfileFieldScreen` but stopped populating the `description` field for new links, causing server-side Mongoose validation (which requires `linksMetadata.description`) to fail when saving profile links.
- Restore a harmless client-side fallback so links created/converted in the UI always satisfy the backend schema and preserve existing UX.

### Description
- Added helper fallbacks `getLinkTitle` and `getLinkDescription` and used them to normalize link items in `packages/services/src/ui/screens/EditProfileFieldScreen.tsx`.
- When initializing list items from `user.linksMetadata` or legacy `user.links`, ensure each item has `title` and `description` before editing.
- When adding a new link and when building the `linksMetadata` save payload, ensure `title` and `description` are present (use fallbacks if missing) so the API receives valid objects.

### Testing
- Verified source-level assertions with a small Node check that the description fallback snippets are present in `EditProfileFieldScreen.tsx` and `git diff --check` returned clean results.
- Attempted `bun run lint`, `bun run typescript`, and `bun run test` in `packages/services`, but those commands are blocked in this environment due to missing local toolchain/dependencies (`biome`, React/RN peers, `jest`), so full type/run tests could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce89f1483239266111a0e4e893d)

### DIFF
--- a/packages/services/src/ui/screens/EditProfileFieldScreen.tsx
+++ b/packages/services/src/ui/screens/EditProfileFieldScreen.tsx
@@ -71,6 +71,9 @@ type EditableListItem = {
     coordinates?: { lat: number; lon: number };
 };
 
+const getLinkTitle = (url: string) => url.replace(/^https?:\/\//, '').replace(/\/$/, '');
+const getLinkDescription = (url: string) => `Link to ${url}`;
+
 /**
  * EditProfileFieldScreen - A dedicated screen for editing profile fields
  *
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #390, opened 2026-06-25
fix: make account data export return Blobs and avoid calling non-existent history endpoint

NateIsern:
### Motivation
- The web data export UI passed the result of `oxyServices.downloadAccountData()` directly to `URL.createObjectURL()`, but the generic HTTP layer parsed `application/json` and `text/csv` responses into objects/strings instead of returning a `Blob`, causing `TypeError` at download time. 
- Activity and Location history UI actions invoked a generic `clearUserHistory()` that issued `DELETE /history`, but the API in this repository does not mount a generic `/history` endpoint and scoped deletion endpoints do not exist, so the UI was performing a nonfunctional/destructive action.

### Description
- Add a `responseType?: 'blob'` option to the core `RequestOptions` and handle `config.responseType === 'blob'` in `HttpService` by calling `response.blob()` before JSON/text parsing so callers can explicitly request raw `Blob` bodies. (`packages/core/src/HttpService.ts`)
- Request account data exports with `responseType: 'blob'` in `downloadAccountData()` so `/users/me/data?format=` responses are returned as `Blob`s and can be consumed by `URL.createObjectURL()` or native sharing. (`packages/core/src/mixins/OxyServices.user.ts`)
- Replace the destructive history-clear flow in the accounts DataScreen with an informational alert until scoped API endpoints for activity/location deletion exist, preventing accidental calls to a non-mounted endpoint. (`packages/accounts/app/(tabs)/data.tsx`)
- Add English and Spanish copy keys for the informational messages. (`packages/accounts/lib/i18n/locales/en.json`, `packages/accounts/lib/i18n/locales/es.json`)

### Testing
- Ran `bun install` to prepare the workspace; this failed due to registry `403` errors when fetching package tarballs, so dependency-driven checks were blocked. (failed)
- Ran lint in `packages/core` via the repository script (`bun run lint`); it could not run because development tooling (`biome`) and deps were unavailable. (failed / blocked)
- Ran TypeScript checks in `packages/core` (`bun run typescript`); this was blocked by missing dependencies/types after the failed install. (failed / blocked)
- Attempted package tests (`bun run test` / Jest) in `packages/accounts`; `jest` was not available due to the install failure, so tests could not be executed. (failed / blocked)
- Ran `git diff --check` and basic repository checks after changes; no diff-check issues were reported. (succeeded)

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cbce8ca988323b39aed48fff43325)

### DIFF
--- a/packages/accounts/app/(tabs)/data.tsx
+++ b/packages/accounts/app/(tabs)/data.tsx
@@ -201,37 +201,16 @@ export default function DataScreen() {
     },
   ], [colors, dataSharing, locationSharing, analyticsSharing, showActivity, handlePrivacyUpdate, pendingPrivacyKey, t]);
 
-  // Handle clear history
-  const handleClearHistory = useCallback(async (type: 'activity' | 'location') => {
-    const title = type === 'activity' ? t('data.activity.clearActivityTitle') : t('data.activity.clearLocationTitle');
+  // Handle clear history. Keep this as an informational action until the API
+  // exposes scoped deletion endpoints for activity and location history.
+  const handleClearHistory = useCallback((type: 'activity' | 'location') => {
+    const title = type === 'activity' ? t('data.activity.activityHistory') : t('data.activity.locationHistory');
     const message = type === 'activity'
-      ? t('data.activity.clearActivityMessage')
-      : t('data.activity.clearLocationMessage');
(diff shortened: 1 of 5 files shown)


NateIsern:

PR #393, opened 2026-06-25
Allow mailbox-only email searches (backend validation)

NateIsern:
### Motivation
- The inbox UI started submitting mailbox-only searches (e.g. `in:spam`) but the API `/email/search` controller still rejected requests that did not include q/from/to/subject/hasAttachment/dateAfter/dateBefore, causing a functional regression; the backend must accept `mailbox` as a valid search criterion so mailbox-only queries reach the search service.

### Description
- Include `mailboxId` in the controller's required-criteria check in `packages/api/src/controllers/email.controller.ts` so `mailbox` counts as a valid filter and mailbox-only searches are allowed to proceed to `emailService.searchMessages`.
- Add a controller unit test `packages/api/src/controllers/__tests__/email.controller.search.test.ts` that verifies mailbox-only searches call `emailService.searchMessages` with the expected args and that requests with no criteria are still rejected.
- Preserve existing call-site behavior so `emailService.searchMessages` receives the same parameters (`q || ''`, `limit`, `offset`, `mailboxId`, `from`, `to`, `subject`, `hasAttachment`, `dateAfter`, `dateBefore`).

### Testing
- Added a Jest unit test file `packages/api/src/controllers/__tests__/email.controller.search.test.ts` covering mailbox-only success and the no-criteria rejection; the test is included in the PR.
- Attempted to run `bun run test -- email.controller.search.test.ts`, but execution was blocked because `jest` was not available in the environment and `bun install` failed due to registry 403 responses, so automated tests could not be executed here.
- Performed static/code inspection to confirm the controller now treats `mailbox` as a valid filter and that the service call arguments are unchanged.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d18bae388832399f8573c4a82a512)

### DIFF
--- a/packages/api/src/controllers/__tests__/email.controller.search.test.ts
+++ b/packages/api/src/controllers/__tests__/email.controller.search.test.ts
@@ -1,0 +1,97 @@
+import type { Response } from 'express';
+
+const mockSearchMessages = jest.fn();
+
+jest.mock('../../services/email.service', () => ({
+  emailService: {
+    searchMessages: (...args: unknown[]) => mockSearchMessages(...args),
+  },
+}));
+
+jest.mock('../../services/smtp.outbound', () => ({
+  smtpOutbound: {},
+}));
+
+jest.mock('../../services/assetServiceSingleton', () => ({
+  assetService: {},
+}));
+
+jest.mock('../../config/email.config', () => ({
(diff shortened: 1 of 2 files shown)


NateIsern:

PR #395, opened 2026-06-25
fix: preserve inbox export failures

NateIsern:
### Motivation
- The export wrapper could mask real `expo export` failures when a stale `packages/inbox/dist/index.html` existed, so failed builds might be treated as successful during CI/deploy.

### Description
- Update `packages/inbox/scripts/export-web.js` so a normal child-process exit only succeeds when the process exit `code === 0` (the previous `existsSync(distPath)` check is limited to the timeout/hang path), and correct the timeout message to match the 8-minute limit.

### Testing
- Performed a syntax check with `node --check packages/inbox/scripts/export-web.js` and ran a targeted fake-`bun` invocation where `bun x expo export --platform web` exited `42` while a stale `packages/inbox/dist/index.html` existed, and confirmed the wrapper now exits with status `1` (test passed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d18ba82fc8323b3a446adf1bd69ee)

### DIFF
--- a/packages/inbox/scripts/export-web.js
+++ b/packages/inbox/scripts/export-web.js
@@ -58,8 +58,9 @@ child.stderr.on('data', (data) => {
 
 child.on('exit', (code) => {
   if (!done) {
-    // Process exited on its own (no hang) — check if it succeeded
-    if (code === 0 || existsSync(distPath)) {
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #397, opened 2026-06-25
fix: fail accounts export on stale dist artifacts

NateIsern:
### Motivation
- Prevent the export wrapper in `packages/accounts` from masking failed or timed-out `expo export` runs when a stale `packages/accounts/dist/index.html` already exists, ensuring build/release integrity.

### Description
- Removed the stale-artifact checks in `packages/accounts/scripts/export-web.js` so the wrapper now treats the child process exit code as the authoritative result instead of using `dist/index.html` presence to force success.
- Adjusted the timeout branch to fail after the configured hard limit (8 minutes) and preserved the original Metro-hang workaround that still succeeds when `Exported:` is observed on stdout and then kills the hung process group.

### Testing
- Simulated a failing `expo export` by placing a fake `bun` in `PATH` that exits nonzero while a stale `packages/accounts/dist/index.html` exists, and verified the wrapper now exits with status `1` as expected. (Succeeded)
- Ran a Node syntax check with `node --check packages/accounts/scripts/export-web.js` to validate the script (Succeeded).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d18b9c7388323a23906f8bdbc83ee)

### DIFF
--- a/packages/accounts/scripts/export-web.js
+++ b/packages/accounts/scripts/export-web.js
@@ -10,13 +10,10 @@
  */
 
 const { spawn } = require('child_process');
-const { existsSync } = require('fs');
 const path = require('path');
 
 const EXPORT_TIMEOUT_MS = 8 * 60 * 1000; // 8 minutes hard limit (DO build instances are slow)
 const projectRoot = path.resolve(__dirname, '..');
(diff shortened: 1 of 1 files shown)


NateIsern:

PR #402, opened 2026-06-25
fix(auth): disclose OAuth consent permissions

NateIsern:
### Motivation
- The redesigned consent UI stopped displaying requested OAuth scopes and app provenance while the backend continued to accept and persist the `scope` query value, which could allow apps to obtain broader permissions than users were shown. 

### Description
- Parse and deduplicate the URL `scope` parameter with `parseRequestedScopes(scope, application.scopes)` in `packages/auth/src/pages/authorize.tsx` and pass the resolved `requestedScopes` into the consent component. 
- Extend `ConsentCard` (`packages/auth/components/consent-card.tsx`) to accept `requestedScopes`, render app provenance/website info, and list friendly permission labels using `labelForScope`. 
- Add localized copy keys for provenance and permissions to `packages/auth/lib/i18n/locales/en.ts` and `packages/auth/lib/i18n/locales/es.ts` and use those keys in the consent card. 
- The changes preserve the existing IdP decision flow and still delegate approve/deny actions to the unchanged `onDecision` handler; only the presentation now discloses the concrete scopes being requested (falling back to the app's registered `application.scopes` when no `scope` parameter is present). 

### Testing
- Ran `git diff --check` which reported no whitespace/patch errors and passed. 
- Attempted to run `bun run --cwd packages/auth build` but the local build is blocked because `vite` is not available in the environment. 
- Attempted `bun install` to restore dependencies but the registry returned HTTP 403 for required packages, so an environment build/test run could not be completed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d15f8c0bc83238a56d1e022d65808)

### DIFF
--- a/packages/auth/components/consent-card.tsx
+++ b/packages/auth/components/consent-card.tsx
@@ -1,5 +1,11 @@
 import { Link } from "react-router-dom";
-import { KeyRound, RefreshCw, ShieldCheck } from "lucide-react";
+import {
+  BadgeCheck,
+  Globe,
+  KeyRound,
+  RefreshCw,
+  ShieldCheck,
+} from "lucide-react";
 import { getAccountDisplayName } from "@oxyhq/core";
 import type { PublicApplication } from "@oxyhq/core";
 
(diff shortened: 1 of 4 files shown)


NateIsern:
✅ Landed on main (wave-2). auth tsc clean; component tests green.

PR #413, opened 2026-06-25
fix(api): allow asset uploads after deleted sha tombstones

NateIsern:
### Motivation
- A recent change excluded `deleted` tombstones from SHA-256 dedup lookups but left a global unique index on `File.sha256`, which caused `E11000` duplicate-key failures when a tombstone with the same hash exists and blocked fresh uploads of identical content.  
- The intent is to preserve deduplication for live assets while allowing new inserts when only a deleted tombstone remains, restoring availability without reintroducing the prior revive/reassign ownership path.

### Description
- Removed the field-level global `unique: true` constraint from the `sha256` field and added a named partial unique index that applies only to non-deleted files (`status: 'active' | 'trash'`) via `FileSchema.index(... { partialFilterExpression: ... })` in `packages/api/src/models/File.ts`.  
- Added a migration helper script `packages/api/scripts/migrate-file-sha256-index.ts` and a `migrate:file-sha256-index` npm script to drop any legacy global `sha256` index and ensure the new partial unique index exists in deployed databases.  
- Added regression coverage `packages/api/src/models/__tests__/File.indexes.test.ts` asserting the schema exposes the partial unique `sha256` index and does not define a global unique `sha256` index.  
- Kept existing duplicate-race recovery and tombstone-exclusion behavior intact so uploads that collide with active records remain deduped while tombstoned hashes no longer block new inserts.

### Testing
- Ran a local source assertion that verified the `File` schema no longer contains the field-level global `unique: true` and that the named partial unique index `sha256_not_deleted_unique` with `partialFilterExpression` for `active`/`trash` is present, which succeeded.  
- Added a unit test `packages/api/src/models/__tests__/File.indexes.test.ts` to assert the new index shape (test file present but full `jest` execution could not be run in this environment).  
- Attempted to run package tests via `bun run --cwd packages/api test ...` but `jest` and other dependencies were unavailable in the execution environment, and `bun install --frozen-lockfile` failed due to registry 403 errors, so full automated test runs could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d49fa41a883238261e8836cb3246f)

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -16,7 +16,8 @@
     "lint:fix": "eslint . --fix",
     "openapi:generate": "bun scripts/generate-openapi.ts",
     "prepublishOnly": "bun run clean && bun run build",
-    "release": "rm -rf dist && bun run build && release-it"
+    "release": "rm -rf dist && bun run build && release-it",
+    "migrate:file-sha256-index": "bun scripts/migrate-file-sha256-index.ts"
   },
   "release-it": {
     "git": {
--- a/packages/api/scripts/migrate-file-sha256-index.ts
+++ b/packages/api/scripts/migrate-file-sha256-index.ts
@@ -1,0 +1,48 @@
+import dotenv from 'dotenv';
+import mongoose from 'mongoose';
+import { File } from '../src/models/File';
+
+dotenv.config();
+
+async function main(): Promise<void> {
+  const uri = process.env.MONGODB_URI;
+  if (!uri) {
+    throw new Error('MONGODB_URI is required');
+  }
+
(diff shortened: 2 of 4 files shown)


NateIsern:
Superseded by #414. #414 lands the same File.ts live-only partial unique sha256 index plus an on-connect migration (`ensureFileSha256LiveUniqueIndex`) that drops the legacy global `sha256_1` unique index, and it also carries the assetService dedup-security changes. The standalone migration script approach here is redundant with #414's on-connect migration. Closing in favor of the complete fix.

PR #420, opened 2026-06-28
fix(api): require origin proof for trusted session clients

NateIsern:
### Motivation

- Prevent trusted first-party/internal applications with public client IDs from being abused to start unbound no-Origin device flows that yield secret `sessionToken` and approval payloads. 
- The recent registration of official public clients (Commons / Oxy Auth) made them selectable by mere `clientId`, and the existing flow treated absence of `Origin`/`Referer` as a native-client signal, enabling server-side attackers to initiate phishing flows. 
- The change is intended to harden `/auth/session/create` so official-branded sessions require origin proof.

### Description

- Require a registered origin for trusted applications in the `/auth/session/create` handler by enforcing the origin check when `isTrustedApplication(resolvedApp)` is true, instead of skipping it when `Origin`/`Referer` are absent; the relevant logic is in `packages/api/src/routes/auth.ts` (bound-origin enforcement). 
- Remove the no-origin/native-client bypass that previously allowed trusted apps to be accepted without an `Origin` header, preventing server-side scripts from starting official-branded device flows. 
- Update the regression test `packages/api/src/routes/__tests__/authSessionAppIdentity.test.ts` to expect rejection (403) for official apps submitted with neither `Origin` nor `Referer`.

### Testing

- Ran `git diff --check` to validate the patch formatting and found no CI lint errors. 
- Attempted to run the focused test `bun run test authSessionAppIdentity.test.ts` but execution was blocked because `jest`/test dependencies are not installed in the environment. 
- Attempted `bun install --frozen-lockfile` to reproduce CI but it failed due to registry tarball `403` responses, so full automated test execution could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3e544886d08323adbf0c6abfb80621)

### DIFF
--- a/packages/api/src/routes/__tests__/authSessionAppIdentity.test.ts
+++ b/packages/api/src/routes/__tests__/authSessionAppIdentity.test.ts
@@ -387,10 +387,10 @@ describe('POST /auth/session/create — application resolution (#214)', () => {
     expect(mockAuthSessionCreate).not.toHaveBeenCalled();
   });
 
-  it('(b4) accepts an official app from a native client that carries no Origin/Referer', async () => {
-    // Native (Expo deviceFlowSignIn) requests attach neither Origin nor Referer.
-    // They cannot prove an origin and must NOT be rejected for lacking one — the
-    // device-flow consent screen still authorises every session interactively.
+  it('(b4) rejects an official app from a no-Origin/no-Referer request', async () => {
+    // Absence of browser headers is not a client authenticator. Server-side
+    // scripts can omit both Origin and Referer, so official public client IDs
+    // must still prove a registered origin before receiving official branding.
     mockApplicationCredentialFindOne.mockResolvedValueOnce(usableCredential(OFFICIAL_APP_ID));
     mockApplicationFindById.mockResolvedValueOnce(officialApp());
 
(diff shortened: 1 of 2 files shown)


NateIsern:
Holding this one for a maintainer decision rather than merging it in the security-PR cleanup batch.

The change makes **every** trusted/official application require a registered-origin proof on `POST /auth/session/create`, removing the `hasBrowserContext(req)` exemption. The problem: native clients (Expo `deviceFlowSignIn`) attach **no** `Origin`/`Referer` header, and `isTrustedApplication()` is true for all official/first-party apps (Commons, accounts, inbox, etc.). With this change those native official-app device sign-ins would get `403 'Application origin is not allowed'` — a regression of the documented, intentional behavior in the `(b4)` test this PR inverts.

The spoofing vector it closes (a server-side script presenting a public official `client_id` with no Origin to get 'official branding' on the consent screen) still requires the user to **interactively approve** the session on the Commons consent UI, so the security upside is limited while the native-sign-in breakage is severe.

To land this safely we'd need a native-client authentication story (e.g. attestation, or treating the interactive device-consent as the authorization for no-origin native flows) rather than a blanket origin requirement. Deferring to you on the preferred direction.

NateIsern:
Superseded — not merging this approach. Removing the no-Origin native bypass in POST /auth/session/create breaks ALL native 'Sign in with Oxy' for official apps (RN sends no Origin → 403 for every trusted app), and it's also incomplete (the same device-code consent-phishing works via a third-party lookalike app that is never origin-gated). The device flow's security boundary is the on-device key signature, but that's exactly what an attacker phishes the victim into producing, so origin-gating /session/create doesn't close the lure.

Proper fix (in progress): keep native allowed; add an authoritative `originVerified` flag (server computes `isTrustedApplication(app) && boundOrigin && applicationAllowsOrigin(app, boundOrigin)`, persists it on the AuthSession, returns it from /auth/session/approve-info), and Commons renders a non-dismissible phishing warning + de-emphasised branding when `originVerified === false`. That covers native + third-party + no-Origin-script without breaking native sign-in.

PR #519, opened 2026-07-03
feat(auth): direct IdP sign-in joins the browser's central device session

NateIsern:
Signing in at auth.oxy.so with another account left it OFF the browser's DeviceSession → diverging account lists across apps (user-reproduced). Login/signup/2FA accept an optional priorSessionId (the SPA already holds it from the account chooser); the server re-validates it (active-only) and derives the deviceId — a raw client deviceId is never trusted (device-fixation guard) — then threads it into createSession and registers the account via the same addAccount+broadcast path as /session/device/add. Response shapes unchanged; missing/invalid prior = today's behavior. api 1337 · IdP 152 · tsc 0 · builds green.

### DIFF
--- a/packages/api/src/controllers/__tests__/sessionAccess.controller.test.ts
+++ b/packages/api/src/controllers/__tests__/sessionAccess.controller.test.ts
@@ -107,6 +107,18 @@ jest.mock('../../services/signature.service', () => ({
   default: {},
 }));
 
+// session.controller imports deviceSession.service for the central device-join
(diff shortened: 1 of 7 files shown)


NateIsern:
Superseded by the auth-platform replanteo (Fases 0-7, PRs #556-#561) and the incoming 2c transport: the IdP now mounts OxyProvider with coldBoot={false} (no session authority), the device-account chooser is already device-first, and 2c removes the browser-central device cookie this PR was built around. Nothing here survives the new architecture.

PR #526, opened 2026-07-04
feat(api): device-first auth surface — /auth/device/* bootstrap, deviceToken, rotating refresh, signout family revoke

NateIsern:
## Phase F2 — SERVER ADDITIVE (auth centralization, wave 1)

**Additive only.** The legacy surface (`/sso*`, `/fedcm/*`, `/auth/refresh`, `/auth/refresh-all`, `POST /auth/session`, and the `oxy_rt` cookie machinery in `refreshToken.service`) is **byte-untouched**. Everything here is new files or additive edits. The api Jest suite is the regression fence: **152 suites / 1492 tests green** (`tsc --noEmit` clean, `eslint` clean, no dependency/lockfile change).

### New endpoints (`src/routes/deviceAuth.ts`, mounted at `/auth` before the generic `/auth`)

| Endpoint | Auth | Function |
|---|---|---|
| `GET /auth/device/bootstrap?return_to&state` | `oxy_device` cookie | Top-level hop: (re-)plants cookie, resolves the active session, mints a single-use origin-bound boot code, 303s back with `#oxy_boot=<fragment>` (no token / no deviceId in the URL). `return_to` is https-only (http loopback exempt), credential-free, ≤2KB, origin must pass `isTrustedOrigin` (third-party never). `Cache-Control: no-store` + `Referrer-Policy: no-referrer`. |
| `POST /auth/device/web-session` | cookie (same-site) | Fast path for `*.oxy.so`: `requireSameSiteOrigin` + trusted-lane + shared registrable apex → exchanges the cookie for a token bundle, no redirect. |
| `POST /auth/device/exchange` | boot code (GETDEL burn, origin-bound) | Atomic single-use burn → `AuthTokenBundle`. 403 on Origin mismatch, 410 on replay/expired. |
| `POST /auth/refresh-token` | body `refreshToken` | ONE rotating refresh (`rotateRefreshToken`) for web + native. 401 uniform on any failure. |
| `POST /auth/device/token` | bearer | Issues the native-channel device token (deviceId from JWT claims). |
| `POST /auth/device/resolve` | `X-Oxy-Internal` | Device-set feed for the IdP chooser. 404 without the secret; no CORS exposure. |

### New models / services

- **`models/DeviceToken`** — hash-only, bound to `(deviceId, origin, channel)`, 400d sliding expiry.
- **`DeviceSession.cookieKeyHash`** (sparse-unique) + **`AuthSession.deviceId`**.
- **`deviceToken.service`** — issue/resolve/revoke + channel policy (`web` requires an exact `Origin`; `native` requires the `Origin` header ABSENT; never throws).
- **`deviceBootCode.service`** — `ssoCode` pattern, `device:boot:` prefix, 60s TTL, atomic GETDEL, fail-closed, **no token in the stored record**.
- **`deviceLogin.service`** — shared add-only login attribution (lazy-imports the model-heavy services so importing it never forces model eval).
- **`deviceSession.service`** — `addAccount` gains `{activate:'always'|'if-empty'}` (default `always`, existing callers unchanged); `getStateByCookieKey` / `ensureDeviceForCookie`; **signout now cascades `revokeAllFamiliesBySession`** to kill persisted refresh families.
- **`utils/internalSecret`** (`hasValidInternalSecret` extracted from `sso.controller`, pure move) + **`utils/deviceCookie`** (one authoritative `oxy_device` cookie spec).

### Additive login-flow edits

`/auth/login`, `/auth/signup`, `/auth/verify`, `/security/2fa/verify-login`, `/auth/session/create+claim`, `/accounts/:id/switch` accept an optional `deviceToken`; attribution precedence is **oxy_device cookie > deviceToken > none**; the resolved deviceId is threaded into `createSession`; the new session is registered **add-only** (`activate:'if-empty'`, never flips the active account) + broadcast; the response **additively** gains `refreshToken` only on the trusted lane (device binding, trusted-lane Origin, or absent Origin) — third-party-lane origins keep today's shape. `/auth/session/claim` now returns a fresh **rotating** refresh token (not the legacy Session JWT); `/accounts/:id/switch` registers the managed session (`activate:'always'`).

### Security notes

- `deviceToken` is **add-only attribution**: it never reads state and never flips the active account.
- The cookie secret is **not** the deviceId (only `sha256` stored); possessing the cookie reveals nothing.
- Constant-time internal-secret comparison; single-use origin-bound boot codes; no `new Model(req.body)` / `req.body` spread; owner ids resolved server-side.

### Existing tests legitimately extended (justified in the commit)

- `sessionClaim.test` — claim now returns a rotating refresh token; assertion relaxed to any non-empty string + the `oauthCode` mock supplies the hashing helpers now used.
- `deviceSession.service.test` / `accountsSwitch.test` — mock the newly-imported services and assert the new device-set registration.

**Do not merge** — merge = prod deploy; the lead coordinates the merge-train.

### DIFF
--- a/packages/api/src/controllers/session.controller.ts
+++ b/packages/api/src/controllers/session.controller.ts
@@ -26,6 +26,7 @@ import {
   revokeAllUserFamilies,
   clearAllRefreshCookies,
 } from '../services/refreshToken.service';
+import { resolveLoginDeviceId, finalizeDeviceLogin } from '../services/deviceLogin.service';
 import type { AuthRequest } from '../middleware/auth';
 import { exactCaseInsensitiveUsernameRegex } from '../utils/resolveUserIdentifier';
 
@@ -331,7 +332,7 @@ export class SessionController {
    */
   static async signUp(req: Request, res: Response) {
     try {
-      const { email, username, password, name, deviceName, deviceFingerprint } = req.body;
+      const { email, username, password, name, deviceName, deviceFingerprint, deviceToken } = req.body;
 
       if (
         !email ||
@@ -412,15 +413,32 @@ export class SessionController {
         });
       }
 
+      // Device-first attribution (oxy_device cookie > deviceToken > none),
+      // resolved before mint so the session carries the central deviceId.
+      const signupDeviceId = await resolveLoginDeviceId(req, deviceToken);
+
       const session = await sessionService.createSession(
         user._id.toString(),
         req,
-        { deviceName, deviceFingerprint }
+        { deviceName, deviceFingerprint, ...(signupDeviceId ? { deviceId: signupDeviceId } : {}) }
       );
 
-      const response = buildSessionAuthResponse(session, user);
-      if (!response) {
+      const baseSignupResponse = buildSessionAuthResponse(session, user);
+      if (!baseSignupResponse) {
         return res.status(500).json({ message: 'Failed to format user data' });
+      }
+      const response: typeof baseSignupResponse & { refreshToken?: string } = baseSignupResponse;
+
+      // Register into the device set (add-only) + broadcast, and additively
+      // attach a rotating refresh token when the lane allows it. Best-effort.
+      const signupDeviceExtras = await finalizeDeviceLogin({
+        req,
+        deviceId: signupDeviceId,
+        session,
+        userId: user._id.toString(),
+      });
+      if (signupDeviceExtras.refreshToken) {
+        response.refreshToken = signupDeviceExtras.refreshToken;
       }
 
       try {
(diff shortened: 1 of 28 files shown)


NateIsern:
### Review findings addressed (commit f3e98ee7)

1. **CodeQL HIGH `js/clear-text-storage-of-sensitive-data` @ deviceCookie.ts (alert #491) — dismissed as false positive (audited).** Data-path audit (grep across `packages/api`): the `oxy_device` value is a random 256-bit CSPRNG bearer-equivalent secret (not user PII). It exists in exactly two places — (a) the `Set-Cookie` transport, hardened `HttpOnly` + `Secure` + `SameSite=Lax`, and (b) server-side **only** as its SHA-256 (`DeviceSession.cookieKeyHash`, sparse-unique). The raw value is never written to any document, cache, or log, and possessing the cookie reveals nothing about the deviceId. There is no cleartext-at-rest to encrypt — the value *is* the credential, exactly like the pre-existing first-party `oxy_rt` refresh cookie (`setRefreshCookie`). The query fires only because this param is named `secret`. Documented inline at `utils/deviceCookie.ts:setDeviceCookie`.
2. **MEDIUM — `assertSameSiteApex` IPv6:** replaced manual `host.split(':')` (shreds `[::1]:3000`) with Express `req.hostname`; kept the registrable-apex match, and for IP-literal / single-label hosts (no apex) fall back to **exact host equality**.
3. **MEDIUM — `/auth/device/resolve`:** accounts now resolve in parallel via `Promise.all`, each isolated in its own try/catch → a dead/corrupt session is skipped (logged), never 500s the whole chooser feed.
4. **LOW — `deviceToken.service`:** `.lean()` on the `DeviceToken` read in `resolveDeviceToken` (read-only, `_id` reused for the bump write).

New tests cover the IPv6/exact-host same-site fallback and the resolve account-skip. Full api suite green: **152 suites / 1495 tests**, `tsc --noEmit` clean, `eslint` clean.

PR #527, opened 2026-07-04
fix(contracts,core): web-session result is a discriminated contract union (session arm carries rotated deviceToken)

NateIsern:
## F2/F3 reconciliation — client-side only (F2's server shape is canonical)

**Server truth (PR #526, verified):** `POST /auth/device/web-session` responds `{data:{reason:'session', session:<AuthTokenBundle>, deviceToken}}` on success and `{data:{reason:'no_session'|'new_device', deviceToken}}` otherwise. After `HttpService.unwrapResponse` strips the `{data}` envelope, the mixin receives the inner object.

F3's client `webSessionResultSchema` was a bare `AuthTokenBundle | no-session` union — it **failed to parse the real success arm** (bundle is nested under `session`, not top-level) **and dropped the rotated `deviceToken`** the server hands back on that arm. This PR aligns the client to the canonical server shape.

### `@oxyhq/contracts` (0.9.1)
- `deviceBoot.ts`: new `webSessionResultSchema` — a `reason`-discriminated union of `WebSessionSession` (`{reason:'session', session: AuthTokenBundle, deviceToken}`) and `WebSessionNoSession` (`{reason:'no_session'|'new_device', deviceToken}`). The nested-session arm is pinned with the explicit-interface + `z.ZodType<>` pattern (node10 `.d.ts` safety, same as `AuthTokenBundle`). Exported `webSessionResultSchema` + types `WebSessionResult` / `WebSessionSession` / `WebSessionNoSession`.
- Tests: both arms + `new_device`, **rejection of the old bare-bundle shape**, missing nested `session`, missing `deviceToken`, unknown `reason`.

### `@oxyhq/core`
- `mixins/OxyServices.deviceBoot.ts`: deleted the local `webSessionResultSchema` / `webSessionNoSessionSchema` / `isAuthTokenBundle` in favor of the contracts import; `requestWebSession` returns the contracts `WebSessionResult`.
- `boot/coldBootV2.ts`: the same-apex web-session step now **persists the rotated `deviceToken` in ALL arms** (it is device-level attribution, not session-level) and uses `result.session` as the bundle on the `session` arm.
- `index.ts`: stopped re-exporting the removed mixin symbols — `WebSessionResult` is owned by `@oxyhq/contracts` and imported from there (core does not re-export contract types).
- Tests: feed the **EXACT post-unwrap server envelope** (the double-unwrap lesson: assert against the real transport shape, not a convenient mock), assert `deviceToken` persistence on the session arm, and assert the old bare-bundle shape is now rejected.

### Verification
- `@oxyhq/contracts`: **152 tests** green (+7). `@oxyhq/core`: **856 tests** green.
- Both builds pass; ESM output verified `require()`-free.
- `bun install` (lockfile committed in the same commit as the 0.9.1 bump) + `bun install --frozen-lockfile` pass. All consumers reference contracts via `workspace:*`/`workspace:^`, so the bump resolves locally with no exact-pin breakage.

### Do NOT merge
Coordinated with the F4/F5 merge train.


NateIsern:
## Expanded scope — 4 bot-review findings folded in (commit b5ebf9a7)

Verified each against the code before implementing; all four were real. Same 0.9.1 bump.

1. **CRITICAL — refresh scheduler tight loop** (`core/src/session/refresh.ts`). Confirmed: an expired / in-lead-window token computed `exp − now − lead <= 0`, `Math.max(fireInMs, 0)` armed `setTimeout(…, 0)`, and a **failing** refresh re-armed at 0 in the finally block → 100%-CPU busy loop (would also blow jest's 100k-timer guard). Fix: a `MIN_SCHEDULE_DELAY_MS` (1s) floor on every delay + exponential failure backoff (5s→…→5min, reset on success/token-change), explicit success/failure branches (no blind finally re-arm), and runRefresh clears any pending timer up front so focus + a fired timer can't double-run. Regression test: expired token + always-failing refresh → bounded (<10) attempts per simulated minute; plus a backoff-reset-on-success test.

2. **HIGH — IP / single-label same-apex** (`core/src/boot/coldBootV2.ts`). Confirmed: last-2-labels collapsed distinct LAN IPs (`192.168.1.1`/`10.0.1.1` → `1.1`) and mishandled IPv6/localhost. Fix: IP literals (v4/v6) and single-label hosts now require **exact** host equality; multi-label hosts keep the registrable-domain compare (co.uk limitation documented; no tldts pulled in). Tests for IPv4/IPv6/localhost.

3. **MEDIUM — storage in-memory mirror** (`core/src/session/authStateStore.ts`). Confirmed: the "stays live in memory" comment had no mirror → private-mode / locked-store sessions silently lost on next `load()`. Fix: both web and native stores keep an in-memory mirror — `save()` writes it first (authoritative even if persist throws), `load()` prefers it over a storage miss/stale, `clear()` nulls it. Tests for quota-throw (web) + locked-keychain (native).

4. **HIGH — fragment code↔reason coupling** (`contracts/src/deviceBoot.ts`). Confirmed: `code` was `.optional()` regardless of reason. Fix: `deviceBootFragmentSchema` is now a `reason`-discriminated union — the `session` arm **requires** `code`; `no_session`/`new_device` omit it. A `session` fragment without a code now fails to parse → treated as "no usable fragment", not half-processed. `core/deviceBootReturn` simplified accordingly (code guaranteed on the session arm).

Nothing refuted — all four reproduced in code.

**Gates:** contracts **154** + core **863** green; both builds pass; ESM `require()`-free; `bun install --frozen-lockfile` clean.

PR #541, opened 2026-07-04
perf(api): GET /apps/authorized bulk $in queries (no N+1); align tests to real query shape

NateIsern:
## Deferred Gemini MEDIUM follow-up (non-blocking) — authorizedApps

Verified the route's REAL query shape first: `GET /apps/authorized` read grants in one `AppGrant.find` but then did a **per-grant** `Application.findById` + `ApplicationCredential.findOne` — an N+1 (2N+1 queries) for a list endpoint. (The prior tests *did* faithfully mock that N+1 chain — not a mock-green≠real-green bug — but the underlying N+1 is the real substance of the finding.)

### Fix
Bulk-resolve applications + their representative active-public `client_id`s in **two `$in` queries**, mapped in memory — **3 queries total, independent of grant count**. Early-return before any join when the user has no grants. Behaviour is identical: same skip-orphaned-grant semantics, same first-active-public-credential-per-app `clientId`.

### Tests
Rewrote the GET tests onto the bulk `Application.find` / `ApplicationCredential.find` shape, asserting **exactly one** bulk call each (proving no per-grant loop) + the `_id: { $in: [...] }` filter + the no-grants short-circuit (no join queries). The DELETE handler is inherently single-key and unchanged.

Full api suite green (145 suites / 1346 tests), `tsc` + `eslint` clean. Only `authorizedApps.ts` + its test changed.

### DIFF
--- a/packages/api/src/routes/__tests__/authorizedApps.test.ts
+++ b/packages/api/src/routes/__tests__/authorizedApps.test.ts
@@ -13,7 +13,8 @@ import { AddressInfo } from 'net';
 
 const mockAppGrantFind = jest.fn();
 const mockAppGrantDeleteOne = jest.fn();
-const mockApplicationFindById = jest.fn();
+const mockApplicationFind = jest.fn();
+const mockCredentialFind = jest.fn();
 const mockCredentialFindOne = jest.fn();
 
 jest.mock('../../middleware/auth', () => ({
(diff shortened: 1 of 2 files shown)


NateIsern:
Obsolete: this optimizes `packages/api/src/routes/authorizedApps.ts`, which no longer exists on `main` — #540 moved authorized-apps onto `AppGrant` and #549 consolidated the surface onto the canonical `/auth/grants` + `connectedApps` mixin, deleting this route. Merging it would resurrect a removed route rather than speed one up. If the N+1 shape came back on `/auth/grants`, that's a fresh PR against the current code.

PR #582, opened 2026-07-08
fix(services): strip device join secrets from URL before first paint

NateIsern:
## Problem

After the device-join redirect from `auth.oxy.so/device/join`, official apps return with credentials in the URL fragment:

```
https://inbox.oxy.so/inbox#oxy_device=…&device_secret=…
```

Those secrets should never linger in the address bar. Previously, stripping only ran asynchronously during cold boot (after React could paint), so users briefly — or persistently, if the bundle loaded late — saw the full credential trace.

## Fix

1. **Sync capture in `OxyProvider`** — on module load, parse the fragment, stage creds in `sessionStorage`, and `history.replaceState` to the clean path before React paints.
2. **Early `<head>` bootstrap** — inline script (exported as `DEVICE_JOIN_URL_STRIP_INLINE_SCRIPT` from `@oxyhq/core`) runs before the JS bundle on inbox, accounts, and console so Expo/Vite code-splitting cannot delay the strip.
3. **`applyDeviceJoinReturn`** reads staged creds from `sessionStorage` first (fallback: parse remaining hash).

## Apps touched

- `@oxyhq/core` / `@oxyhq/services` — capture + strip utilities
- `packages/inbox/app/+html.tsx`
- `packages/accounts/app/+html.tsx`
- `packages/console/index.html`

## Test plan

- [x] `packages/core` deviceJoin tests (11 passed, incl. inline script key parity)
- [ ] After deploy: open accounts → inbox; URL should stay `https://inbox.oxy.so/inbox` with no `#oxy_device` fragment
- [ ] Session sync still works after join return

<div><a href="https://cursor.com/agents/bc-e5350a8a-17ed-46c8-acf9-dfd421014146"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e5350a8a-17ed-46c8-acf9-dfd421014146"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/accounts/app/+html.tsx
+++ b/packages/accounts/app/+html.tsx
@@ -1,4 +1,5 @@
 import { ScrollViewStyleReset } from 'expo-router/html';
+import { DEVICE_JOIN_URL_STRIP_INLINE_SCRIPT } from '@oxyhq/core';
 import type { PropsWithChildren } from 'react';
 
 const DEFAULT_TITLE = 'Accounts by Oxy';
@@ -28,6 +29,9 @@ export default function Root({ children }: PropsWithChildren) {
         <meta property="og:site_name" content="Accounts by Oxy" />
         <meta property="og:title" content={DEFAULT_TITLE} />
         <meta property="og:description" content={DEFAULT_DESCRIPTION} />
+
+        {/* Strip device-join credentials from the URL before the JS bundle loads. */}
+        <script dangerouslySetInnerHTML={{ __html: DEVICE_JOIN_URL_STRIP_INLINE_SCRIPT }} />
 
         <ScrollViewStyleReset />
       </head>
--- a/packages/console/index.html
+++ b/packages/console/index.html
@@ -11,6 +11,10 @@
     <meta name="theme-color" content="#0a0a0a" />
     <meta name="description" content="Oxy Developers Console" />
     <title>%APP_NAME%</title>
+    <!-- Strip device-join credentials before the app bundle loads (mirrors @oxyhq/core). -->
+    <script>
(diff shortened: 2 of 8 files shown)


NateIsern:
Closing: the flow this defends no longer exists.

This strips `oxy_device` / `device_secret` out of the URL fragment left by the `auth.oxy.so/device/join` redirect. On `main` there is no `packages/core/src/utils/deviceJoin.ts` and no `device/join` route anywhere in the tree — the redirect, the `oxy_device` cookie, the `/sso` bounce and the `#oxy_boot` bootstrap were all removed in the device-first (phase 2c) session rework. There is no longer a URL for credentials to land in.

**The finding was correct and the fix was the right shape** — credentials in a fragment are readable by anything that runs before your JS, so stripping them in an inline script *before* the bundle loads, rather than in app code, is the only placement that actually helps. Worth remembering if a redirect ever carries a secret again; it is just not reachable today.

NateIsern:
Closing: the mechanism this patches no longer exists.

`packages/core/src/utils/deviceJoin.ts`, its test, and `packages/services/src/ui/utils/deviceJoin.ts` were all deleted from `main` in the zero-cookie cutover — verified with `git cat-file` against `origin/main`, not inferred. Three of the eight files this PR touches are gone, and they are the three the change is built around; there is no device-join secret in a URL to strip any more. The transport is now a first-party `{deviceId, deviceSecret}` persisted per origin, and OAuth returns either through `postMessage` in popup mode or a single-use `?code=` the SDK exchanges and clears.

Gemini's review comment on it (wrap the `sessionStorage` write and the URL rewrite in `try...catch` so a private-browsing quota or security error cannot leave the app mid-state) is a fair point and worth keeping in mind for any future code that stages a credential in web storage — it just has nothing to attach to here.

Reopen if I have read this wrong.

PR #645, opened 2026-07-16
feat(auth): atomic key rotation + last-credential replacement (b3 F3, backend)

NateIsern:
## b3 Feature 3 — key rotation + last-credential replacement (BACKEND slice)

The security-critical, verifiable-here part (contracts + core + api). The Commons native UI (rotate-key screens) is a **separate follow-up slice** and is NOT in this PR.

Rotation is an **atomic REPLACE** of the single identity key — never remove-then-add — so it never passes through `countAuthMethods() === 0` and is independent of the unlink guards in `authLinking.ts`. Because control of the current key is *proven* (SecureStore in `device` mode, or a recovery-phrase re-derivation in `phrase` mode), even the **last remaining credential** can be replaced.

### contracts (`packages/contracts`)
- `src/keyRotation.ts`: `rotateKeyChallengeResponseSchema`, `rotateKeyCompleteRequestSchema`, `rotateKeyCompleteResponseSchema` + inferred types; exported from `index.ts`. Internal `workspace:*` — **no npm publish in this PR**.

### api (`packages/api`)
- `models/AuthChallenge.ts`: additive `purpose: String` (default `'signin'`) — non-breaking.
- `routes/authLinking.ts`:
  - `POST /auth/rotate/challenge` (bearer, `rl:identity:rotate:challenge:`) — mints a `purpose:'rotate_key'` `AuthChallenge` bound to the account's current key.
  - `POST /auth/rotate/complete` (bearer, `rl:identity:rotate:complete:`) — (1) derives `oldPublicKey` **from the user doc, never the request**; (2) **atomically burns** the purpose-scoped, key-bound challenge; (3) verifies the signature over `JSON.stringify({action:'rotate_key', userId, oldPublicKey, newPublicKey, challenge, timestamp})` against the CURRENT key (same primitive `/auth/link` uses); (4) rejects a `newPublicKey` already registered elsewhere; (5) swaps `user.publicKey` **and** replaces the single `authMethods` identity entry in place (array length unchanged). Invalidates `userCache`. Optional `signOutEverywhere` revokes other sessions. DID needs **zero migration** (derived live).

### core (`packages/core`)
- `crypto/recoveryPhrase.ts`: `derivePendingIdentity()` + `derivePrivateKeyFromPhrase()` — pure, **non-persisting** derivation.
- `mixins/OxyServices.identity.ts`: `rotateKey({ proof, phrase?, signOutEverywhere?, pendingIdentity? })` — derive new identity → sign with old key → challenge/complete → `KeyManager.importKeyPair(overwrite)` **only after server confirmation**. Ambiguous-network-failure guard reconciles against the derived DID (which carries `publicKeyHex`) before surfacing the error.

## Security invariants (proven by tests)
- `oldPublicKey` is always server-derived from the user doc; a client-supplied `oldPublicKey` is ignored and a signature from a different key is rejected (400).
- The `rotate_key` challenge is **purpose-scoped** (a signin challenge can never complete a rotation → 401) and **single-use** (atomic burn).
- Rotation is atomic: `authMethods` length is unchanged (never a `countAuthMethods()===0` window).
- `newPublicKey` already registered → 409; rotating to the same key → 400.
- `userCache.invalidate` fires; the derived DID reflects the new key immediately.
- KeyManager fault injection: fail the primary write *after* the server swap → the old key is still recoverable from backup.

## Verification
- api `tsc --noEmit` clean; full `bun run test` green (**1568** pass, 169 suites).
- core build + tsc + tests green (**891** pass); contracts build + tests green (**133** pass).

## Reviewer notes / follow-ups
- **(a) A `security-reviewer` pass is needed before merge** — atomic swap + last-credential replacement + challenge-burn ordering.
- **(b) The Commons rotate-key UI is the follow-up slice** (not in this PR).
- **(c) `@oxyhq/contracts` needs republish-first at merge time** (before the core/api consumers) so the new schemas are available to any external consumer.
- Deviation from the blueprint's literal wording: the ambiguous-retry reconciliation checks the **derived DID document** rather than `GET /auth/methods`, because the auth-methods contract intentionally does NOT expose raw public keys, whereas the DID's `verificationMethod[].publicKeyHex` is derived live and reflects a completed rotation immediately.

### DIFF
--- a/packages/api/src/controllers/session.controller.ts
+++ b/packages/api/src/controllers/session.controller.ts
@@ -283,11 +283,15 @@ export class SessionController {
         });
       }
 
-      // Find and validate the challenge (read-only query with .lean() for performance)
+      // Find and validate the challenge (read-only query with .lean() for
+      // performance). Scoped to signin-purpose challenges (default, or legacy
+      // docs predating the `purpose` field) so a `rotate_key` challenge cannot be
+      // spent to mint a session.
       const authChallenge = await AuthChallenge.findOne({
         publicKey,
         challenge,
         used: false,
+        purpose: { $in: ['signin', null] },
         expiresAt: { $gt: new Date() }
       }).lean();
 
--- a/packages/api/src/models/AuthChallenge.ts
+++ b/packages/api/src/models/AuthChallenge.ts
@@ -9,6 +9,13 @@ import mongoose, { type Document, Schema } from "mongoose";
 export interface IAuthChallenge extends Document {
   publicKey: string;
   challenge: string;
+  /**
+   * What the challenge may be spent on. Additive (default `'signin'`) so
+   * existing signin flows are unaffected. Purpose-scoped consumers (e.g. key
+   * rotation, which requires `'rotate_key'`) filter on this so a challenge
+   * minted for one flow can never be redeemed by another.
+   */
+  purpose: string;
   expiresAt: Date;
   used: boolean;
   createdAt: Date;
@@ -24,6 +31,10 @@ const AuthChallengeSchema: Schema = new Schema(
       type: String,
       required: true,
       unique: true,
+    },
+    purpose: {
+      type: String,
+      default: 'signin',
     },
     expiresAt: {
       type: Date,
--- a/packages/api/src/models/User.ts
+++ b/packages/api/src/models/User.ts
@@ -459,6 +459,11 @@ const UserSchema: Schema = new Schema(
       unique: true,
       sparse: true,
       trim: true,
+      // Canonical secp256k1 keys are stored uncompressed + lowercased (see
+      // SignatureService.canonicalizePublicKey). `lowercase` guards the case
+      // dimension of the unique index so a re-cased duplicate cannot slip past
+      // it. Existing keys are already lowercase-uncompressed, so this is safe.
+      lowercase: true,
       select: true,
     },
     refreshToken: {
--- a/packages/api/src/routes/__tests__/authLinking.reversibility.test.ts
+++ b/packages/api/src/routes/__tests__/authLinking.reversibility.test.ts
@@ -70,6 +70,20 @@ jest.mock('../../models/WebauthnCredential', () => ({
   },
 }));
 
+// authLinking imports the session service + Session model (for the key-rotation
+// `signOutEverywhere` path); stub them so the real modules — which crash at load
+// under the global mongoose mock (`SessionSchema.methods` is undefined) — are
+// never evaluated. These reversibility tests never exercise rotation.
+jest.mock('../../services/session.service', () => ({
+  __esModule: true,
+  default: { deactivateAllUserSessions: jest.fn() },
+}));
+
+jest.mock('../../models/Session', () => ({
+  __esModule: true,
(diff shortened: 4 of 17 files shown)


NateIsern:
## Security-review fixes applied (commit cbead6e0)

Addressed the MEDIUM + LOWs from the review. No merge — still awaiting sign-off.

### MEDIUM — key re-encoding impersonation (both parts fixed)
1. **Proof-of-possession over `newPublicKey`.** `rotateKeyCompleteRequestSchema` now requires `newKeyProof` — the NEW key signs `{action:'rotate_key_new', userId, newPublicKey, challenge, timestamp}`. The server verifies it against `newPublicKey` before the swap, so an attacker can no longer rotate their account to a re-encoding of a key they don't control. Wired through `rotateKey` (signed with the pending/new key).
2. **Canonicalization.** New `SignatureService.canonicalizePublicKey` (uncompressed, lowercased) is applied before the differ-check, the uniqueness query, AND the write; `User.publicKey` gains `lowercase: true`. The self-check and cross-account check are now both canonical (fixes the case inconsistency), so two encodings of the same point cannot coexist across accounts.

### LOW
- **Symmetric purpose scoping.** The signin/handoff challenge burn now filters `purpose ∈ {signin, null}` in BOTH `authSession.service.ts` (`authorizeSessionWithSignedChallenge`) AND `session.controller.ts` (`verifyChallenge`) — a `rotate_key` challenge can't be spent to mint a session.
- **Timestamp before burn.** The freshness check in `/auth/rotate/complete` now runs BEFORE the atomic burn, so a stale-but-valid request can't self-burn its own challenge.
- **Persist-failure resilience.** `rotateKey` now surfaces the new phrase (with `localPersistFailed: true`) instead of throwing when the on-device persist fails after a server rotation, so the user can re-import the now-live key.

### Tests added
- proof-of-possession required (missing → schema 400; wrong-key proof → 400)
- re-encoded (compressed/uppercased) duplicate rejected + canonical-form stored + uniqueness query runs against the canonical key
- `signin cannot spend a rotate_key challenge` — new `authSessionPurposeScope.test.ts` (rotate_key → 401 at the challenge gate; signin/legacy-null → passes the gate, fails at signature)
- stale request does not burn its challenge
- `localPersistFailed` path returns the phrase without throwing

### Verification
- api `tsc --noEmit` clean; full api build green.
- Full suites green: **contracts 133**, **core 892**, **api 1576**.

### Note
One deviation worth flagging: the new-key proof is verified against the `newPublicKey` value AS SENT (raw), then the key is canonicalized for the uniqueness query + storage. Verifying the raw form is safe (an attacker still cannot produce a valid signature for a key they don't hold, in any encoding), and it keeps the client's signed bytes and the server's reconstruction identical.

NateIsern:
Follow-up (08e6b02b): cleared the CodeQL "database query built from user-controlled sources" alert on the rotate_key challenge burn (`authLinking.ts`) by adding an explicit `typeof challenge !== 'string'` barrier before the Mongo query — mirrors the existing `POST /auth/link` guards; Zod already narrows it, this is defense-in-depth. Full api suite still green (1576).

Issue #691, opened 2026-07-26
feat(auth): fluid popup OAuth with identity-bound Commons approvals


NateIsern:
## Summary

Build a simple, fluid, secure cross-domain **Sign in with Oxy** experience where:

- the relying application stays mounted and never visibly redirects away
- the user performs one primary action
- Oxy automatically selects the best approval route
- Commons remains permanently bound to its local cryptographic identity
- account switching in Mention, Mercaria, Syra, or any other application never changes the identity shown or used by Commons

This issue evolves the existing OAuth, `AuthSession`, Commons signed approval, `DeviceSession`, and `SessionClient` architecture. It must not create a competing authentication system.

### One intention, one primary action, one confirmation

The normal experience must not present a menu of authentication methods.

The user initially sees only:

```text
Continue with Oxy
```

After that click, Oxy chooses the best available route:

- known Commons installation: send the request and wait
- unknown desktop browser: show QR immediately
- mobile browser with Commons available: open Commons directly
- no usable Commons route: reveal a discreet recovery or alternative sign-in path

Do not show multiple equal buttons such as:

- Send notification
- Open Commons
- Scan QR
- Choose device
- Use passkey
- Use password

Fallbacks must remain hidden under a small **Having trouble?** or equivalent affordance until the primary route fails or is unavailable.

### Commons is the identity

Commons is not another relying application and must not behave like one.

Commons represents one local cryptographic identity. That identity is the authority for:

- identity display
- identity synchronization
- signed approvals
- credentials
- identity backup and recovery
- approval notifications

The active account selected elsewhere in the Oxy ecosystem must never replace the Commons identity.

### Identity and account are different concepts

```text
Identity
Who the human cryptographically is
Bound to a local key
Stable inside Commons
```

```text
Account
The context an application is acting as
Personal, organization, project, or bot
Switchable per application or compatible account session
```

Example:

```text
Identity approving the request: Nate
Account Mention will act as: The Oxy Collective
```

Commons remains Nate before, during, and after the approval.

## Current bug: Commons follows the account switcher

Commons currently mounts the generic `OxyProvider`. The provider projects the `DeviceSession.activeAccountId` into the application-level active `user` and active access token.

As a result, switching from a personal account to an organization or project in another application can also change the user shown and used inside Commons.

This is incorrect and must be fixed as part of this work.

[...]

PR #697, opened 2026-07-27
feat(auth): OAuth-bound AuthSession with atomic finalization (#691 phase 3)

NateIsern:
Phase 3 of #691, on top of #696.

## Why

Two disjoint authorization flows existed: the OAuth code flow (`/auth/oauth/authorize` → `AuthCode`) and the Commons signed approval (`AuthSession` + `authorizeCode` + QR). Popup, push, QR and deep link cannot each grow their own business flow, so `AuthSession` is extended to carry an OAuth request and finalize into the standard authorization code. No competing request model; nothing already owned by `Application`, `AuthCode` or `DeviceSession` is duplicated.

The authoritative state machine stays small — `pending → authorized → consumed`, plus `cancelled` / `expired`. Delivery progress is not a status.

## Binding at creation

`POST /auth/session/create` gains optional:

```ts
oauth: { redirectUri; codeChallenge; codeChallengeMethod: 'S256'; scope?; subjectAccountId? }
```

The redirect URI must pass the same exact-match constant-time check the authorize endpoint uses — **403 with no `Location` header** on failure (RFC 6749 §3.1.2.4). Non-`S256` is refused. Scopes are normalized identically. That check now lives in one module both the route and the service import, instead of two copies drifting apart.

## Exactly one code, ever

`finalizedAuthCodeId` is a **reservation, not a post-hoc marker**: the code id is allocated and written in the same conditional update that consumes the session.

```ts
const codeId = new mongoose.Types.ObjectId();
const claimed = await AuthSession.findOneAndUpdate(
  { _id, purpose: 'oauth_authorization', status: 'authorized', finalizedAuthCodeId: null },
  { $set: { finalizedAuthCodeId: codeId, status: 'consumed', consumedAt: now } }, { new: true });
if (!claimed) return { ok: false, reason: 'already_finalized' };
await issueAuthCode({ codeId, ... });   // the one existing code minter, not a second
```

Losing the race, or a later `issueAuthCode` failure, both leave the request spent — the correct fail-closed direction for something that grants access. `POST /auth/session/finalize/:sessionToken` takes no bearer (the secret `sessionToken` is the credential, like `/auth/session/claim`), is rate-limited under `rl:auth:session-finalize:`, and collapses every rejection to `401 invalid_grant` with the reason only in the server log.

## Delegated accounts

A delegated authorization is the identity authorizing a **subject account**, never the identity becoming it. The code is issued for the subject with the approving identity recorded in `AuthCode.operatedByUserId`, and the existing `accountService.verifyActingAs` predicate — the same one `POST /accounts/:id/switch` uses — is re-checked at approval **and** at finalization. A personal account can never be a delegated subject; that would be impersonation. `POST /auth/oauth/token` threads the operator through to `createSession`, so a delegated session is a real managed session bound to live `act_as` membership and dies with revocation.

## One more fail-closed decision

An OAuth-purpose approval mints no session, and `/auth/session/claim` refuses `purpose: 'oauth_authorization'`. Otherwise a surface that only needs a code could pick up an access token, and every approval would strand a live session for the identity.

## SDK

`startCommonsSignIn({ clientId, oauth? })` (additive — the existing call shape is byte-identical), `CommonsApprovalInfo` gains `purpose` + sanitized `subjectAccount`, and `finalizeCommonsOAuth(sessionToken)` returns a **code, never a token**. Parsing is fail-safe in the safe direction: an unrecognised `purpose` degrades to `device_sign_in`, never to "you're granting an OAuth authorization".

## Tests

api 203 suites / 1911 pass (+41), core 98 suites / 1176 pass (+25). Including: unregistered redirect URI → 403 with no redirect; non-S256 refused; approve-info leaks no `sessionToken`/`redirectUri`/`codeChallenge`; **two concurrent finalizes yield exactly one `AuthCode`**, asserted at the model boundary; a finalized session cannot mint a second; the code stays single-use through `exchangeAuthCode` with PKCE; a delegated subject without permission is refused at both approval and finalize; `device_sign_in` sessions unaffected.

### DIFF
--- a/packages/api/src/models/AuthCode.ts
+++ b/packages/api/src/models/AuthCode.ts
@@ -25,8 +25,20 @@ import mongoose, { type Document, Schema } from "mongoose";
 export interface IAuthCode extends Document {
   /** SHA-256 hex digest of the raw authorization code. */
   codeHash: string;
-  /** ObjectId of the user that approved this grant. */
+  /**
+   * SUBJECT of the grant — the account the code authorizes access to. For a
+   * delegated authorization this is the ORGANIZATION/PROJECT account, not the
+   * human that approved it (see `operatedByUserId`).
+   */
   userId: mongoose.Types.ObjectId;
+  /**
+   * OPERATOR identity for a DELEGATED grant: the human identity that approved
+   * the app to act as `userId`. Absent for ordinary self-grants. Modelled
+   * separately so identity and account are never conflated; the session minted
+   * at exchange time carries it as `operatedByUserId`, which binds that
+   * session's validity to the operator's live `account:act_as` membership.
+   */
+  operatedByUserId?: mongoose.Types.ObjectId;
   /** Application ObjectId (string) the code was issued for. */
   appId: string;
   /** Exact redirect URI used at issue time — must match exchange request. */
@@ -62,6 +74,11 @@ const AuthCodeSchema: Schema = new Schema(
       ref: 'User',
       required: true,
       index: true,
+    },
+    operatedByUserId: {
+      type: Schema.Types.ObjectId,
+      ref: 'User',
+      default: null,
     },
     appId: {
       type: String,
--- a/packages/api/src/models/AuthSession.ts
+++ b/packages/api/src/models/AuthSession.ts
@@ -24,8 +24,47 @@ import mongoose, { type Document, Schema } from "mongoose";
  *                   `applicationId` at create time) — there is no free-form app
  *                   label. It links the session to authoritative, sanitized app
  *                   metadata (name/icon/badge/scopes) for the consent UI.
+ *
+ * Purpose:
+ *   A session is EITHER a device sign-in (the historical QR/app-to-app handoff,
+ *   claimed for an access token via `POST /auth/session/claim`) OR an OAuth
+ *   authorization request (finalized into a single-use `AuthCode` via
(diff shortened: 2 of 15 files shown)


NateIsern:
CodeQL note for the record.

The check reports 2 new alerts. Only one was actually introduced here:

- **#540 `js/insufficient-password-hash` → `oauthCode.service.ts:28`** — genuinely new (this PR adds a call to `sha256Hex` from `finalizeOAuthAuthorization`), and a false positive: it hashes a 32-byte CSPRNG authorization code at rest (`AUTH_CODE_BYTES = 32`), not a user password. A 256-bit random credential is not brute-forceable, and a slow KDF would add latency to every token exchange for no gain. Dismissed with that reasoning.
- **#404 (`auth.ts:1509`) and #338 (`auth.ts:652`) `js/sql-injection`** — pre-existing on `main` since 2026-06-15 and 2026-02-06 respectively, at those exact lines. They are re-attributed to this PR only because the diff touches the same file. I checked them rather than waving them through: both are exact-match lookups on validated input — `Application.findById(credential.applicationId)` (id read from the resolved DB credential) and `Application.findById(applicationId)` behind `isValidObjectId`, plus `authorizeCode` from `validate({ params: authorizeCodeParams })` used as a string equality. No object ever reaches a query position, so there is no NoSQL injection here. Left open — they are not this PR's to dismiss.

PR #711, opened 2026-07-27
chore(deps): Express 5 — groundwork (NOT ready to merge)

NateIsern:
Draft on purpose. Replaces the blind merge of #433 with the parts that are
actually safe, and measures what the real migration costs.

## What this branch already fixes
- `express@^5.2.1` + `@types/express@^5.0.6` in api/node; dev-only in protocol
  and federation (their peers already allowed `^5`).
- **`@oxyhq/core`'s published peer widened** `^4.0.0` → `^4.18.0 || ^5.0.0`
  rather than replaced — replacing it would break every consumer still on
  express 4 (Mention, Homiio, Allo backends).
- Root `overrides` pin `@types/express` + `@types/express-serve-static-core` to
  v5, so `@types/multer` / `compression` / `swagger-ui-express` /
  `express-slow-down` stop dragging nested v4 copies. Two structurally identical
  `Request` types is what produced the duplicate-identity errors.
- Dropped `@types/express-rate-limit` — deprecated stub; v8 ships its own types.
- **Two real runtime breakages fixed** (these would have hit prod, not CI):
  - `middleware/validate.ts`: `req.query` is getter-only in Express 5, so the
    plain assignment throws on EVERY validated route. Now redefines the property.
  - three rate-limit tests: `app.all('*')` → `'/*splat'` for path-to-regexp v8.
  - `protocol/node/nodeApp.ts`: narrow a route param before `.toLowerCase()`.

## What remains — and why it isn't in here
**187 type errors, all one shape.** `@types/express-serve-static-core@5` declares
`ParamsDictionary[key: string]: string | string[]` (v5 + path-to-regexp v8 can
bind array params), so every `req.params.x` used as a string needs narrowing —
~30 files across `packages/api`. It cannot be fixed globally: `ParamsDictionary`
is an interface, and declaration merging can add to an index signature but not
narrow it.

Doing that mechanically right now is a bad trade:
- it is ~187 edits on the API that fronts all of prod, and CI is the only net;
- `packages/api` is being actively edited by other work in flight, so a
  30-file sweep would conflict head-on;
- Express 4.22.x is still maintained, and nothing is blocked on 5.

**Good news for whoever picks this up:** the production router is clean. A repo
scan found NO `*`, optional-`?`, or regex route patterns outside those three
test files, and no `res.send(status)` / `req.param()` / `res.redirect('back')` /
`app.del()` usage. path-to-regexp v8 — normally the scary part of this
migration — is a non-event here.

Closes nothing yet; #433 stays open as dependabot's tracker.


NateIsern:
Revalidated against current main: this draft was explicitly incomplete and current API/Docker guards intentionally pin Express 4 and its single type identity. Closing the stale groundwork rather than merging an unverified major-runtime migration. A future Express 5 cutover needs its own current branch and full DB/API suite.

PR #717, opened 2026-07-28
fix(seed): stop clobbering application capabilities, and bound a run

NateIsern:
## The bug

`seed-oxy-applications.ts` put `capabilities: []` in `desiredAppFields` and assigned it authoritatively on every reconcile. `SeedAppSpec` has no `capabilities` field, so that empty array was a default rather than a declaration.

A re-run would therefore set `Commons by Oxy`'s capabilities to `[]`, stripping the `identity:approval` grant that `register-commons-clients.ts` unions on. `authSessionDelivery.service.ts` selects push-delivery targets by exactly that capability, so pending sign-in requests would end up with zero eligible devices — and nothing would report an error.

`DRY_RUN` could not have warned about it. For an existing application the reconcile branch is `else if (!dryRun)`, so a dry run computes nothing and reports `appsUpdated: 0` regardless of what a real run would change.

## Bounding a run

`SEED_ONLY_APPS` restricts a run to named applications. It resolves before the database connection, so a misspelling costs nothing and never reaches the database, and an unknown name is refused rather than skipped — a run that reconciles nothing and exits 0 is indistinguishable from a successful one.

`redirectUris` remains a destructive replace. That is defensible as a declarative canonical list, but it means a full run reverts any out-of-band redirect URI on every other application, which is a second reason to bound a production run.

## CrowdSource

Added so its reviewer app at `https://crowdsource.oxy.so` can sign in. Sign-in only, so it takes the default `['user:read']`.

It must never carry `reputation:write`: a device sign-in's granted scopes **are** the application's scopes, so adding it would widen every device sign-in rather than only a service credential — and that scope also confers opening a civic validation jury, not just ledger writes.

### DIFF
--- a/packages/api/scripts/seed-oxy-applications.ts
+++ b/packages/api/scripts/seed-oxy-applications.ts
@@ -238,6 +238,24 @@ const SEED_APPS: SeedAppSpec[] = [
       'https://hub.moovo.now',
     ],
   },
+  {
+    name: 'CrowdSource',
+    description:
+      'Official Oxy participatory moderation platform — reviewers are assigned cases by the server and review them blind.',
+    websiteUrl: 'https://crowdsource.oxy.so',
+    type: 'first_party',
+    // The reviewer app ships to the web only today (Cloudflare Pages project
+    // `crowdsource-frontend`). Its Expo `crowdsource://` deep-link scheme is
+    // registered here when a native build actually ships — an unused redirect
+    // surface is authority nobody asked for.
+    redirectUris: ['https://crowdsource.oxy.so'],
+    // Sign-in ONLY, so the default `['user:read']`. CrowdSource must never
+    // carry `reputation:write`: it never moves an Oxy Trust figure itself, it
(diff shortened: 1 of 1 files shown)


NateIsern:
Closing in favour of #718, which is the better implementation of the same two changes.

#718 puts the selection decision in a pure module under `src/` — so it is covered by the package tsconfig and unit-testable, following the existing `registerCommonsClientsPlan.ts` precedent — and it ships 17 tests with both guards mutation-tested. This PR's filter lived inline in `scripts/`, which is outside the tsconfig include and effectively untestable.

It also fixes a real flaw in this one: here, `SEED_ONLY_APPS=""` parses to zero names and falls through to the full seventeen-application run. An operator who meant to bound the run and passed a variable that happened to be empty would get exactly the blast radius the filter exists to prevent. #718 throws on a set-but-empty value.

Note the env var name differs — #718 uses `ONLY_APPS`. Whichever invocation is documented must match whichever implementation lands, or the filter is silently ignored.

PR #728, opened 2026-07-29
fix(services)!: make the five core expo peers required, not optional

NateIsern:
Follow-up to #723, which fixed `expo-notifications` by moving the push adapter behind an export subpath and flagged these five as the remaining exposure. Same defect class, opposite correct fix.

## The defect

Five modules are named from code reachable from the root barrel while being declared in a way that does not install them. Verified against the **published `23.0.0` tarball**, not the repo:

| module | declared in 23.0.0 | its own devDep | reachable in tsc's graph | reachable in Metro's graph |
|---|---|---|---|---|
| `expo-image-picker` | **optional** peer | ❌ | ✅ | ✅ |
| `expo-image-manipulator` | **optional** peer | ❌ | ✅ | ✅ |
| `expo-haptics` | **optional** peer | ✅ | ✅ | ✅ |
| `expo-web-browser` | **not declared at all** | ❌ | ✅ | ✅ |
| `expo-document-picker` | **optional** peer | ❌ | ❌ (latent) | ✅ |

`tsc` resolves the specifier of an `import()` even when the call is lazy and wrapped in try/catch, and a package manager does not install an optional peer. Against a consumer harness with the five genuinely absent, the published shape fails exactly:

```
@oxyhq/services/src/ui/screens/ChangeAvatarScreen.tsx(121,35): error TS2307: Cannot find module 'expo-image-picker'
@oxyhq/services/src/ui/screens/AvatarCropScreen.tsx(184,35):   error TS2307: Cannot find module 'expo-image-manipulator'
@oxyhq/services/src/ui/screens/AvatarCropScreen.tsx(285,39):   error TS2307: Cannot find module 'expo-haptics'
@oxyhq/services/src/ui/components/oauthNavigation.ts(69,35):   error TS2307: Cannot find module 'expo-web-browser'
@oxyhq/bloom/src/hooks/useHaptics.tsx(10,52):                  error TS2307: Cannot find module 'expo-haptics'
```

`expo-web-browser` is the worst of the five: it is not a peer, not a dev dependency, and has no stub. It resolved only because every app that happens to use services also installs it.

## Two module graphs, and why `expo-document-picker` differs

`ui/navigation/routes.ts` loads screens through `require()`. `tsc` does not follow `require()`; Metro does. So the barrel's reachable set is not one set but two, and the five do not sit in the same place:

- `ChangeAvatarScreen` and `AvatarCropScreen` are pulled into a consumer's **type** program by `import type` edges (`surfaceRegistry.ts`, `useAvatarPicker.ts`), so their specifiers are live TS2307s today. Confirmed with `--listFiles`: both files are in the program.
- `FileManagementScreen` is reached **only** through `require()`, so `expo-document-picker` breaks nobody's type-check right now. It is still a broken core surface: `handleFileUpload` is the only "add files" action in the file manager on every platform, and it throws `expo-document-picker is not installed` without it. It is also one static import away from becoming a TS2307 like the others.

## Verdicts — decided per module from the code, not applied wholesale

The test #723 set is whether the module has an internal consumer. None of these five is a notifications-shaped case; all five have one, and all five back core SDK surfaces, so the honest fix is reclassification rather than more subpaths. Hiding `ChangeAvatar` / `AvatarCrop` / `FileManagement` / the native OAuth lane behind subpaths would move core UI out of the SDK's own router to protect a declaration.

| module | verdict | why |
|---|---|---|
| `expo-image-picker` | **required peer** | `ChangeAvatarScreen`, the SDK's `ChangeAvatar` route. Both device sources depend on it. |
| `expo-image-manipulator` | **required peer** | `AvatarCropScreen`. Its loader *throws* when absent — the crop editor is the only path to setting an avatar. AGENTS.md already called it required in prose. |
| `expo-haptics` | **required peer** | Same barrel chain. Functionally graceful (returns `null`), but there is no shape that keeps it optional *and* keeps `AvatarCrop` in the barrel. Every RN consumer already installs it, and Bloom — a required peer of this package — names it too. |
| `expo-web-browser` | **required peer** | `oauthNavigation.ts`, the third-party "Sign in with Oxy" native lane. Degrades to `Linking.openURL`, but must at minimum be *declared*. |
| `expo-document-picker` | **required peer** | `FileManagementScreen`'s only upload path. Not a tsc break today; a complete runtime break of a core route, and a latent tsc break. |

All five ship web implementations, so this is not pushing native-only modules onto web consumers — and the package already requires `expo-image`, `expo-file-system`, `@expo/vector-icons`, `react-native-reanimated` and `react-native-svg` from those same consumers.

### Rejected: an ambient `declare module`

Same reason #723 rejected it — it shadows the real package for the entire consumer program. Worse here: `src/types/expo-haptics.d.ts` and `src/types/expo-document-picker.d.ts` already existed and were shadowing the real `expo-haptics` inside this package's own build while consumers still broke. **Both are deleted.** `expo-crypto.d.ts` / `expo-secure-store.d.ts` stay: those modules are loaded through a runtime-computed specifier (`const SECURE_STORE_MODULE = 'expo-secure-store'`), deliberately invisible to both resolvers.

## The package's own type-check was passing by accident — confirmed

Installed from its own manifest alone, into an empty directory:

```
before:  332 errors  (@oxyhq/bloom ×331, expo-image-picker, expo-image-manipulator, expo-web-browser)
after:     0 errors  (210 source files loaded)
```

`@oxyhq/bloom`, `expo-image-picker`, `expo-image-manipulator`, `expo-web-browser` and `@testing-library/react` were resolving purely from sibling workspace packages hoisting them into the root `node_modules`. All are now this package's own devDependencies, along with `react` and `react-native` (pinned to the versions the rest of the monorepo already uses), so the rule has no exceptions to allowlist.

> One trap worth recording: the first isolation run reported **1 error** and looked like a pass. A broken implicit `@types` entry (`TS2688: Cannot find type definition file for 'parse-path'`) aborts `tsc` before it reports a single source error. The harness now sets `types: []` and asserts a floor of ≥200 loaded files before trusting the result.

## Verification — all three resolvers, against a harness where the modules are genuinely absent

Consumer harness mirroring CrowdSource's reviewer dependency set, with all five **removed from its manifest**, using the packed tarball from this branch.

| resolver | before | after |
|---|---|---|
| `tsc`, `react-native` condition (`customConditions: ["react-native"]` → published `src/`) | ❌ 5 errors | ✅ **0 errors**, 135 services source files in the program |
| `tsc`, `import`/bundler condition (Vite/web → `lib/**/*.d.ts`) | ✅ 0 (masked by `skipLibCheck`) | ✅ 0 errors |
| Metro, five installed | ✅ 2473 modules | ✅ 2473 modules |
| Metro, five `rm -rf`'d from `node_modules` | ✅ 2455 modules | ✅ **2455 modules, exit 0** |

The install itself is the fix's mechanism and is directly observable — the harness manifest never lists the five:

```
expo-image-picker: installed        expo-haptics:      installed
expo-document-picker: installed     expo-web-browser:  installed
expo-image-manipulator: installed   expo-notifications: NOT installed   <- still optional; the differential is caused by this change
```

The 18-module Metro differential confirms the dynamic imports genuinely resolve when installed rather than being silently dropped, and the absent run proves the runtime shape is unchanged: graceful degradation, no new build failure. Mutation-tested — adding a *static* `import * as x from 'expo-web-browser'` to `oauthNavigation.ts` fails the Metro build with `Unable to resolve module expo-web-browser from …/oauthNavigation.ts`, so the Metro check can distinguish success from failure. Run with a dedicated `METRO_CACHE_ROOT` and `--clear`.

## Regression gate

`packages/services/__tests__/packaging/peerDeclarations.test.ts` walks the **real module graph** in two variants — tsc's (imports only) and Metro's (also `require()`) — and asserts that nothing reachable in either names a module a consumer is not guaranteed to have. A third assertion keeps the isolation fix from rotting: every module named anywhere in `src/` must be in this package's own `dependencies`/`devDependencies`.

Vacuity floors: ≥100 tsc-reachable files, ≥150 Metro-reachable, Metro strictly greater than tsc, ≥180 scanned files naming ≥20 external modules, plus known-file anchors asserting `FileManagementScreen` is in the Metro graph and *not* in the tsc graph.

Mutation-tested, five ways:

| mutation | result |
|---|---|
| `expo-image-picker` → optional | ✕ both graph assertions, printing `index.ts -> ui/navigation/bottomSheetManager.ts -> ui/navigation/surfaceRegistry.ts -> ui/screens/ChangeAvatarScreen.tsx` |
| `expo-web-browser` removed from peers | ✕ both, `expo-web-browser [UNDECLARED] is named by:` |
| `expo-document-picker` → optional | ✓ tsc assertion, ✕ **Metro assertion only** — the two graphs are genuinely distinguished, not duplicated |
| `expo-image-picker` + `@oxyhq/bloom` dropped from devDeps | ✕ isolation assertion, naming both and their namers |
| traversal collapsed (`EXTENSIONS = ['.bogus']`) | ✕ 3 vacuity assertions; the offender assertions pass vacuously, which is exactly what the floors exist to catch |

#723's `barrelIsolation.test.ts` stays scoped to the two specifiers the push adapter owns so it still fails for one reason only; its comment claiming these five are "legitimately reachable ... optional" is updated, since that is now the wrong half of the sentence.

## Tests

Each via the package's own command, never a blanket `bun test`.

| package | result |
|---|---|
| `@oxyhq/services` | 63 suites, **554 passed** (62 / 547 on `origin/main`, + the 7 new gate assertions) |
| `commons` | 62 suites, **561 passed** |
| `accounts` | 14 suites, **136 passed** |
| `inbox` | 3 suites, **26 passed** |

`bun run typescript`: clean. `bun run lint`: **8 pre-existing errors** (6 `noArrayIndexKey`, 2 `useExhaustiveDependencies`), byte-identical to `origin/main` measured on a stashed working tree — none in a file this PR touches. `bun.lock` updated in this commit and verified in sync.

## Version impact — and a correction on the release state

**No call sites change, in this repo or any other.** No exported API moves. Consumers get the modules from their next `bun install`.

It is still semver-major: a consumer that refuses the install now has an unmet required peer where it previously had a satisfied optional one.

**The `23.0.0` on npm does not contain #723's fix.** It was published 2026-07-29T03:12Z from `a34dc472` (#724, the #691 phase-7b release), which is an ancestor of #723's merge `c790dc44`. The published tarball still re-exports the adapter from `src/index.ts` and has no `./notifications` entry in its `exports` map. So the notifications break is unreleased, that version number is burned, and **both breaking changes should ship together as `24.0.0`** rather than as two majors in a row.

Newly installed per consumer on the next install (nothing to edit):

| consumer | newly installed |
|---|---|
| Mention frontend, Homiio frontend, Alia app, Syra frontend | none — already has all five |
| Allo frontend, CrowdSource reviewer | `expo-document-picker` |
| Syra studio | `expo-image-manipulator` |
| Syra sdk | `expo-document-picker`, `expo-image-manipulator`, `expo-haptics`, `expo-web-browser` |
| in-repo `accounts`, `commons` | `expo-image-picker` |
| in-repo `inbox` | `expo-image-manipulator` |
| in-repo `oxy-console`, `auth`, `test-app-vite` (web-only) | all five |

## Two findings outside this PR's scope

1. **Bloom has the identical defect.** `@oxyhq/bloom` declares `expo-haptics` as an optional peer `*` and names it from `src/hooks/useHaptics.tsx`, which is why that fifth error appears in the baseline above. Requiring it here masks the symptom for services' consumers; a direct Bloom consumer still hits it. Belongs in Bloom.
2. **The `>=56.0.0` peer ranges now bite.** With these peers required, a package manager actually resolves them — and `>=56.0.0` let bun pull SDK-57 modules into the SDK-56 harness. Harmless while they were optional and never installed. There is already a `fix/services-sdk56-peer-ranges` branch in flight, so I left the ranges alone rather than collide with it.

Not merged, not published, no version bumped — release-it owns that.


NateIsern:
Partially landed, and the remaining live piece is now in #770.

**What is already on `main`:** four of the five are required peers — `expo-image-picker`, `expo-image-manipulator`, `expo-haptics`, `expo-document-picker` — and `optionalPeers` is gone from `packages/services/package.json` entirely.

**What was still missing, and #770 fixes:** `expo-web-browser` was declared in *no* dependency field at all, while `oauthNavigation.ts:69` does `await import("expo-web-browser")` and both `OxySignInButton` and `OxyContext` reach it from the root barrel. #770 declares it and adds `barrelSpecifiersDeclared.test.ts` — a module-graph walk with a vacuity floor, mutation-verified — so the class cannot recur silently.

**Leaving this open**, because two substantial pieces are neither landed nor in #770 and are worth doing on their own terms:

- the devDependencies that only type-checked by hoisting accident (**332 errors installed standalone** — that measurement is the valuable part, and the empty-directory + `types: []` harness is the right way to get it);
- deleting the ambient `declare module` stubs that SHADOW the real package for a consumer that did install the module.

The `peerDeclarations.test.ts` here is also more thorough than what #770 ships — two graph variants, `tsc`'s and Metro's — and the observation that a specifier reachable only through `require()` still backs a screen the SDK router can open is the kind of thing nobody rediscovers. Worth rebasing rather than abandoning.

NateIsern:
@dependabot rebase

NateIsern:
Triaged against `main` @ `b5cef38d`. **Premise live, diff stale — it needs re-authoring, not rebasing.**

The finding still holds exactly. All five are still declared optional on `main`:

```
expo-haptics           optional=true
expo-document-picker   optional=true
expo-image-manipulator optional=true
expo-image-picker      optional=true
expo-web-browser       optional=true
```

so `@oxyhq/services` still tells a consumer these are droppable while its own code imports them.

What blocks the rebase is everything else this diff carries from 2026-07-29:

- **`"@oxyhq/bloom": "^0.67.0"` in devDependencies.** `main` is on catalog `^1.9.2` with a `^1.0.0` peer; this would take the dev install back across the 1.0 boundary. Same reason #970 was just closed.
- **Hard-pinned versions where `main` now uses catalog references.** This writes `expo-document-picker: ~57.0.1`, `expo-image-picker: ~57.0.4`, `react: 19.2.3`, `react-native: 0.86.0`; `main` resolves `expo-web-browser` (and its neighbours) through `catalog:`. Pinning them here reintroduces the per-package drift the catalog exists to prevent.
- It also touches `AGENTS.md` and `bun.lock`, both long since moved on.

One thing worth being precise about before doing the work: under bun an unsatisfied peer **warns and never nests**, so flipping `optional` off does not make a consumer's install fail — it changes what the package *declares*, and the enforcement has to come from the consumer's own install check. That is still worth having (the declaration is currently false), but it is a correctness fix rather than a guard, and the `!` in the title promises consumers more breakage than they will actually see.

The re-authored version is small: delete the five `peerDependenciesMeta` entries, leave every version reference on `main`'s catalog, touch nothing else.

NateIsern:
Validated again on current main: making these peers required forces a fresh resolution that is currently blocked by the repository minimum-release-age policy for Expo 57, so it is not safely installable today. Closing rather than merging a non-reproducible lockfile; the change remains recoverable.

PR #750, opened 2026-07-31
fix(docker): narrowing the workspace subset must not drop the catalog

NateIsern:
`Deploy to AWS` failed on the first push after #749, at `bun install`:

```
error: elliptic@catalog: failed to resolve
error: express@catalog: failed to resolve
error: helmet@catalog: failed to resolve      (…and seven more)
ERROR: process "/bin/sh -c bun install" did not complete successfully
```

Nothing was deployed — the build never produced an image, so the running service is still the previous one.

## Cause

Both image builds copy the monorepo root `package.json` and then rewrite `workspaces` to just the packages that image needs:

```js
p.workspaces = ['packages/contracts', 'packages/protocol', …]
```

Written as an array, that assignment also deletes `workspaces.catalog`. Every `"catalog:"` reference in the copied manifests then has nothing to resolve against — and these builds install *without a lockfile* on purpose ("workspace subset doesn't match the full monorepo lock"), so there is no fallback.

## Fix

Write the subset as the object form, carrying the catalog through:

```js
p.workspaces = { packages: [...], catalog: (p.workspaces && p.workspaces.catalog) || {} }
```

which is all that assignment ever meant to do — narrow the package list.

`packages/node/Dockerfile` has the same two rewrites and the same latent break. No workflow builds it today, so it is fixed here rather than found later.

## Verification

- `docker build` of the full API image (both stages) — succeeds
- control: the same build with `origin/main`'s Dockerfile against this tree — fails with the exact `catalog: failed to resolve` errors from the CI log

That control is the point: it shows this is the fix, not a coincidence.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,11 @@ WORKDIR /app
 # Remove bun.lock since the workspace change invalidates it — bun will
 # resolve fresh dependencies (still deterministic from package.json versions).
 COPY package.json ./
-RUN node -e "const p=require('./package.json'); p.workspaces=['packages/contracts','packages/protocol','packages/federation','packages/core','packages/api']; delete p.patchedDependencies; require('fs').writeFileSync('package.json', JSON.stringify(p, null, 2));"
+# The subset is written as the OBJECT form on purpose: narrowing `workspaces`
+# to an array drops `workspaces.catalog` with it, and every `"catalog:"`
+# reference in these manifests then fails to resolve — `bun install` here has
+# no lockfile to fall back on, so the build dies at dependency resolution.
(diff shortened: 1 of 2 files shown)


NateIsern:
Superseded by e1c778d6, which landed the equivalent fix directly on main while this was in CI and covers the same four rewrites (api ×2, node ×2). `Deploy to AWS` is green on that commit, so the API deploys again.

Leaving one measurement here that the landed commit does not carry: the failure reproduces locally with `docker build` against origin/main's Dockerfile (`elliptic@catalog: failed to resolve`, ten packages) and the fixed one builds the full image green — the control that shows the fix is the fix. Worth knowing that this path is reproducible without waiting for CI, since these builds install with no lockfile and so have no fallback when a manifest reference cannot resolve.

PR #752, opened 2026-07-31
fix(api): stop deleting subscriptions on expiry, derive entitlement from endDate

NateIsern:
## Summary

`models/Subscription.ts` declared a Mongo TTL index — `SubscriptionSchema.index({ endDate: 1 }, { expireAfterSeconds: 0 })`. A Mongo TTL index **deletes** the document once the indexed field is in the past; it does not mark it. The enum already had a `'expired'` status value, so the author plainly meant to transition the row, not destroy it.

The practical effect: every subscription row was deleted the moment its `endDate` passed.

- No subscription history survives past its lapse — no record a user ever held a plan, no data for billing disputes, renewals, or analytics.
- `autoRenew: true` rows were deleted instead of renewed.
- `status: 'expired'` was unreachable — nothing had ever transitioned a row to it, because the row was gone before anything could.
- A lapsed `pro` subscriber silently dropped to free-tier storage limits (`storage.controller.ts`) with zero audit trail of what happened or when.

**Already-deleted history is largely unrecoverable.** The daily `mongodump` only contains a row that still existed at the moment it ran — anything that lapsed before the oldest retained snapshot is in no backup at all. `BillingSubscription` (the Stripe-era collection) was never touched by this index — that billing history is intact. This bug only affected the legacy `Subscription` collection.

## What this change does

1. **Removes the TTL index** and replaces it with two regular indexes (`{status,endDate}`, `{userId,status,endDate}}`) that serve the entitlement lookup and the reconciliation sweep. `models/__tests__/subscription.indexes.test.ts` fails if a TTL index reappears on this schema.
2. **Derives entitlement from `endDate` vs now, never from the stored `status`.** This part was not optional: `resolveUserSubscriptionPlan` filtered on `status: 'active'`, and nothing in the codebase ever wrote `status: 'expired'` — the TTL deletion was the *only* mechanism that had ever terminated paid entitlement. Removing the index alone, without this, would have converted a data-loss bug into "paid features forever" for every already-lapsed row. `utils/subscriptionStatus.ts` is the single definition of "is this subscription in force right now," shared by the entitlement check, the response serializer, and the owner-facing controller.
3. **Keeps `status` accurate for reporting** via lazy expiry-on-read (`findCurrentSubscription`) plus an `unref()`'d 15-minute sweep (`subscriptionLifecycle.service.ts`), mirroring the existing `DevicePairingSession` pattern. Neither is load-bearing for entitlement — a missed sweep tick only delays a reporting value, never grants or revokes access.
4. **Adds `packages/api/scripts/drop-subscription-ttl-index.ts`** (one-shot, `DRY_RUN=1` by default) and **`.github/workflows/run-drop-subscription-ttl-index.yml`** (`workflow_dispatch`, `dry_run: true` default) to drop the now-undeclared but still-live index in production. Removing the schema declaration does **not** drop a live index — Mongoose's `autoIndex` only ever *creates* declared indexes, it never drops undeclared ones — so until that workflow runs, production keeps deleting rows even with this fix deployed.

## Deploy order

Code first, index drop second. Dropping the index before this code ships would leave already-lapsed rows sitting at `status: 'active'` indefinitely, granting `pro` forever under the old status-only check.

**The index-drop workflow is not being run as part of this PR.** Nate is approving and running it separately after this merges and deploys.

## Test plan

- [x] `bun run test` (packages/api) — 238/238 suites, 2333/2333 tests passing, including new coverage: `subscription.indexes.test.ts` (regression guard against a TTL index reappearing), `subscriptionStatus.test.ts`, `subscriptionLifecycle.service.test.ts`
- [x] `bunx tsc --noEmit` (packages/api) — clean
- [x] `bun run build` (packages/api) — clean
- [x] `bun install` + `git diff --exit-code -- bun.lock` — no drift (no package.json touched by this change)
- [ ] Deploy, then separately run `run-drop-subscription-ttl-index.yml` with `dry_run: true` first, verify its output names exactly one TTL index on `endDate`, then re-run with `dry_run: false`

### DIFF
--- a/.github/workflows/run-drop-subscription-ttl-index.yml
+++ b/.github/workflows/run-drop-subscription-ttl-index.yml
@@ -1,0 +1,183 @@
+name: Drop the subscription TTL index
+
+# One-shot: drop the TTL index on `subscriptions.endDate`, which DELETED every
+# subscription row the moment its period ended (see
+# packages/api/scripts/drop-subscription-ttl-index.ts for the full story).
+#
+# Removing the declaration from the schema does NOT drop the live index —
+# Mongoose's `autoIndex` only ever CREATES declared indexes — so until this runs,
+# production keeps deleting rows even with the fixed code deployed.
+#
+# WHY IT IS SHAPED LIKE THIS (same reasoning as run-user-text-backfill.yml):
+#
+# 1. The database is on a PRIVATE VPC address, unreachable from a GitHub runner,
+#    so the script has to run as an ECS task INSIDE the VPC ("Fargate one-shot").
+#
+# 2. It reuses the LIVE service's task definition (same secrets, role, subnets,
+#    security groups) and only overrides the image + command, touching nothing the
+#    service runs. Building from `ref` lets it run from a branch before it merges.
+#
+# The script touches index METADATA only — it reads, writes and deletes no
+# document — and drops only an index on `{ endDate: 1 }` that actually carries
+# `expireAfterSeconds`. It is idempotent: a second run reports "already absent".
+#
+# ALWAYS run dry FIRST (`dry_run=true`, the default): it lists every index on the
+# collection and names exactly what it would drop, changing nothing.
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Branch or SHA to build the one-shot image from'
+        required: true
+        type: string
+        default: 'main'
+      dry_run:
+        description: 'Report what WOULD be dropped, change nothing'
+        required: true
+        type: boolean
+        default: true
+
+concurrency:
+  # Never let two index operations touch the collection at once.
+  group: drop-subscription-ttl-index
+  cancel-in-progress: false
+
+env:
+  AWS_REGION: us-west-2
+  APP: oxy-api
+  CLUSTER: oxy-cluster
+  SERVICE: oxy-api
+  DOCKERFILE: Dockerfile
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  drop-index:
+    # Same arch as the deployed image — the service runs linux/arm64.
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 60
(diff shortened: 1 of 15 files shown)


NateIsern:
Closing: the bug is fixed on `main`, by the Drizzle/PostgreSQL port (`eed7a183`), and fixed **better** than this branch proposed.

**The finding was right and serious.** `SubscriptionSchema.index({ endDate: 1 }, { expireAfterSeconds: 0 })` is a TTL index: it DELETES the row when the period closes, so billing history is destroyed and "never subscribed" becomes indistinguishable from "subscribed and lapsed".

**But the premise no longer holds.** Nothing imports `models/Subscription` on `main` any more — `subscription.controller.ts` is entirely on Drizzle (`getDb()`, `db/schema/subscriptions`). The Mongoose model is dead code, so its declared index is never created by the live app, and this branch's controller/util changes are written against a layer that no longer exists.

**`db/schema/subscriptions.ts` on `main` addresses it head-on**, and its comment reads like it was written with this branch in view:

- it names the Mongo TTL index and what it destroyed;
- the table is **deliberately not registered in `db/expiry.ts`**, because that registry deletes rows;
- expiry is **derived** — `status <> 'canceled' and end_date <= now()` — with `status = 'expired'` materialized by a projection job for reads, rather than being the trigger for a delete;
- it even records that Mongo's TTL monitor lagged ~60s, so an expired subscription stayed readable for a while — the kind of detail only someone who had actually looked at this would know.

Deriving the state instead of destroying the row is the right shape, and it is what this branch was reaching for.

**One thing worth checking separately, which neither addresses:** whether the live MongoDB still carries that index on a `subscriptions` collection. Removing a declaration never drops a live index — Mongoose's `autoIndex` only ever CREATES — which is exactly the point this branch made and is worth keeping. It is now data hygiene on a collection nothing reads rather than active data loss, but the one-shot script here is a reasonable starting point if anyone wants to confirm it.

Issue #759, opened 2026-08-02
Video mp4 renditions (360p/720p/1080p) 404 on production files despite upload-time generation, and cannot be lazily regenerated


NateIsern:
## What I measured

Probing a real production video file (`6a390d4b9d8fecf98a320181`, h264 **720x1280**, 6,016,175 B, public + active) across the whole video variant taxonomy:

```
?variant=poster      -> 302   (real JPEG frame)
?variant=hls_master  -> 302
?variant=360p        -> 404
?variant=720p        -> 404
?variant=1080p       -> 404
```

`poster` and `hls_master` resolving means upload-time `VariantService.generateVideoVariants` **did run** for this file — those two are produced by the same pass that produces the mp4 renditions. Yet no rendition resolves.

Two candidate explanations, and I did not disambiguate them (it needs a look at the `File.variants` array in prod, or at the S3 keys):

1. `generateVideoVariant` failed per-rendition and returned `null`. It deliberately swallows the failure (`resolve(null)` on non-zero exit, on a spawn error, and on an upload error) so one bad rendition doesn't abort the whole pass — which also means a silent, total rendition loss looks identical to a healthy upload from the outside.
2. The variant rows exist but their S3 objects don't, so `ensureVariant`'s `s3Service.fileExists` probe drops them and falls through.

`360p` should not have been skipped by the size guard: the skip is `config.width > metadata.width`, and 640 ≤ 720.

## Why it can't self-heal

`assetService.ensureVariant` can lazily regenerate an image variant, and (as of #758) a video's poster and sized poster renders. It has **no lazy path for `360p`/`720p`/`1080p`** — those are produced only by the upload-time pass, so a file that missed them at upload never gets them.

## Impact

Low today, which is why this is an issue rather than a fix. Mention plays video via `hls_master` plus the raw original URL and never requests a rendition, and the survey of the other Oxy apps found no rendition consumer either. The concern is what it implies: **the upload-time video path may be failing more often than it appears**, and the swallow-and-continue design means nothing surfaces when it does. `hls_master` resolving here is reassuring but not conclusive — HLS is generated by a separate `generateHLSStream` call with its own error handling.

## Suggested next steps

- Query prod for the `variants[].type` distribution across `mime: /^video\//` files — how many have `poster`/`hls_master` but no rendition? That single query separates "rare" from "systemic" and distinguishes the two explanations above.
[...]

PR #762, opened 2026-08-02
fix(assets): actually queue variant generation instead of running it inline

NateIsern:
## The bug

`AssetService.queueVariantGeneration` was a method named `queue…` that queued nothing — it awaited `variantService.generateVariants` directly.

**Correction to the original diagnosis:** all eight call sites are floating promises, so the HTTP *response* was never blocked. What was unbounded was the **concurrency**. Every upload immediately started up to seven sharp encodes (image) or three x264 transcodes plus HLS segmentation (video), in-process, with nothing limiting how many ran at once.

On a 512 CPU / 1024 MiB Fargate task that pins the container, and the JS thread then loses the CPU long enough to miss the ELB's 5 s `/health` probe three times running — so the load balancer kills a task that is merely *starved*, mid-upload, and the work is lost. Measured: service-average CPU ~85% sustained for six hours, per-task memory at exactly 1023/1024 MiB, `Task failed ELB health checks` as recently as `2026-08-02T07:10:30Z`.

## The change

Generation moves to `queue/assetVariants.queue.ts`, following the existing `linkPreviewWarm.queue.ts` shape:

- **BullMQ** when `REDIS_URL` is set — one job per file, deduped by a stable colon-free `av-<sha256(fileId)>` id, worker concurrency **1**, 3 attempts with exponential backoff.
- **No Redis** → a deduped **sequential** in-process drain. Deliberately *not* a fallback to inline generation: inline is the bug, so reproducing it wherever Redis is absent would reintroduce the outage in the environment least able to absorb it. One file at a time is still strictly better-bounded than the unbounded fan-out it replaces.
- The enqueue never throws into the caller — an upload that already stored its bytes must not fail on a queue outage.

### Design decisions

| Question | Decision |
|---|---|
| What does a request return before a variant exists? | **Unchanged.** `GET /cdn/:id?variant=` calls `ensureVariant`, which *materialises the variant on demand* for image sizes, posters, and (since `64ef2459`) mp4 renditions. No new 404 for those. |
| Does generation stay lazy-on-read? | **Yes — it is the correctness backstop.** The queued job pre-warms; the lazy path covers the gap while a job waits. They cannot duplicate work: variant keys are content-addressed (same `sha256`+type → same key), and the commit no longer clobbers. |
| Idempotency / dedup | BullMQ ignores an `add` for a jobId already queued/active. The fallback dedupes on a pending/in-flight set. |
| Failure behaviour | A failed generation now **throws**, marking the job failed and retried, instead of being swallowed. |
| Redis absent | Sequential drain, above — never inline. |

### Also fixed

- **#759 — swallowed total loss.** `generateVideoVariant` resolves `null` on ffmpeg failure, so a video whose *every* rendition failed reached the end of the method, logged "generated successfully", and was indistinguishable from a healthy upload. Now raised — counting attempts separately from successes, since a source smaller than every target legitimately attempts nothing. Raised **after** the commit so the poster that did encode is still persisted.
- **`replaceVariants` → `upsertVariantSet`**, scoping the delete to the types being written. The whole-set clear destroyed renditions the lazy read path had just materialised. Harmless when generation started immediately; a growing window once a job can sit in a queue.
- **`s3ServiceSingleton` extracted.** The queue module is imported *by* `assetService.ts` and needs storage; `assetServiceSingleton` constructs an `AssetService` at module-eval time, so reaching it from there resolves the half-initialised class to `undefined` and throws at boot. Only 2 files imported `s3Service`.

## One consumer-visible timing note

`hls_master` is the one variant with **no lazy read path** (`ensureVariant` handles image sizes, `poster`, and mp4 renditions; an HLS master falls through and 404s). Queueing therefore lengthens the window in which a freshly-uploaded video has no HLS playlist.

This is a window that already existed and that the consumer is explicitly built for — Mention's `mediaResolver.ts` says so in situ:

> Adaptive-bitrate HLS master playlist. **NOT guaranteed to exist yet** (background transcode is fire-and-forget on upload) — the frontend player **MUST fall back to `url`** (the raw original) on a playback error.

So this lengthens an anticipated gap rather than creating a new failure mode. Worth knowing when reviewing, and worth revisiting if the queue ever runs a deep backlog.

## Verification

- `bun run test` — **4386 passed / 4386, 304 suites**
- `bun run lint` — exit 0 (92 pre-existing warnings, unchanged count; none in changed files)
- `bun run build` — **3 `error TS`, all three inherited from `main`** (see below); this PR adds zero.

**Mutation-tested**, restoring the file byte-identical after each (md5 verified):

| Mutation | Result |
|---|---|
| inline `void this.variantService.generateVariants(file.id)` | `does not run variant generation during the request` fails — *"Received number of calls: 1"* |
| the above **plus** an awaited call site (the originally-described bug) | `resolves without waiting for generation to finish` **times out at 10 s** — the request-path-blocked signature |
| whole-set delete restored in `upsertVariantSet` | `preserves rows of types the batch does not write` fails |

### What the tests caught in *this* PR

Both of these were defects in my own first cut, found by tests rather than by review — the clearest evidence the suite here is load-bearing:

1. **A hung generation permanently wedged the fallback drain.** `draining` stayed `true` forever when a generation never settled (a wedged ffmpeg is the realistic case), so the in-process path would silently never process another file again — the opposite failure from the unbounded one this PR fixes, and just as total. Fixed with a start/stop epoch that retires a parked drain.
2. **Module-eval-time construction broke an unrelated suite.** The queue module built its `VariantService` at import; since it sits in the import graph of anything touching `assetService`, that ran `new VariantService(...)` before the importing test's own top-level `const` spies existed → TDZ `ReferenceError` at import in `assetService.ensureVariantVideo.test.ts`. Now constructed on first use.

### Known gap

**The all-renditions-failed raise has no test.** Reaching it requires `child_process.spawn` mocked three different ways (ffprobe, poster frame, rendition) and no such machinery exists in this repo — disproportionate for a single assertion. Flagged here so a reviewer sees the gap rather than discovering it.

## `API Build` is red from `main`, not from this PR

`64ef2459` (*fix(api): lazy mp4 rendition regeneration*) introduced three type errors in `variantService.ts`; `main` has been red on `API Build` since, including HEAD `8bc503cd`. Measured on the same tree: `origin/main` alone → **3** `error TS`; this branch rebased onto it → **3**. `git blame` puts all three on that commit. Not fixed here — it is another agent's in-flight area.

## Notes

- Pairs with the task resize in `oxy-infra` ([#18](https://github.com/OxyHQ/oxy-infra/pull/18)). This PR bounds the work; that one gives it room.
- The worker runs on the API tasks. A dedicated worker service would be the next step if variant volume keeps growing — deliberately not built here.
- Measured and ruled out: `sharp.concurrency()` already defaults to **1**, so libvips is not spawning host-CPU-count threads.
- **Correction to an earlier claim of mine: the HLS pipeline is NOT unconsumed.** I had grepped only this monorepo and wrongly generalised. Mention's `packages/backend/src/utils/mediaResolver.ts` requests `hls_master` for every native video and serves it as `hlsUrl` (live in prod on `GET /feed/mtn?descriptor=videos`), feeding hls.js over MSE because desktop Chrome and Firefox do not decode HLS natively. **Do not delete the HLS pipeline.** The mp4 renditions (`360p`/`720p`/`1080p`) have no *named* consumer in any Oxy checkout outside this repo's own tests — but `64ef2459` just added a lazy on-read path for them, so they are reachable via `?variant=720p` without any repo naming the literal. Neither is delete-ready.


NateIsern:
### Rebased onto current `main` — and a heads-up on `API Build`

**`API Build` fails on this PR for a reason that is not this PR.** It is inherited from `main`.

`64ef2459` (*fix(api): lazy mp4 rendition regeneration*) introduced three type errors in `variantService.ts`, and `main` has been red on `API Build` ever since — including its current HEAD `8bc503cd`:

```
src/services/variantService.ts(1414,18): error TS2339: Property 'width' does not exist on type '{}'.
src/services/variantService.ts(1414,67): error TS2339: Property 'width' does not exist on type '{}'.
src/services/variantService.ts(1416,71): error TS2339: Property 'width' does not exist on type '{}'.
```

Same three errors here, same three columns, offset to 1444/1446 by this PR's additions above them. `git blame` puts all of them on that commit.

Measured both sides on the same tree:

| tree | `bun run build` `error TS` count |
|---|---|
| `origin/main` alone | **3** |
| this branch (rebased onto it) | **3** |

This PR adds **zero**. The offending line is `metadata.width` where `metadata` comes from `file.metadata?.video ?? await this.extractVideoMetadataFromUrl(...)` — `file.metadata?.video` widens to `{}`. Not fixed here: it is another agent's in-flight area and this PR should not carry it.

`API Build` on this PR cannot go green until `main` is fixed.

### Verification after the rebase

Re-ran everything against the rebased tree (fresh `bun install`, dedicated PostGIS):

```
Tests:       4386 passed, 4386 total
Test Suites: 304 passed, 304 total
✖ 92 problems (0 errors, 92 warnings)     LINT_EXIT=0
```

That 304th suite is `assetService.ensureVariantVideo.test.ts`, which arrived with `64ef2459` — and it caught a real defect in this PR's first cut. The queue module built its `VariantService` at module-evaluation time; since it sits in the import graph of anything touching `assetService`, that ran `new VariantService(...)` before the importing test's own top-level `const` spies existed, giving a TDZ `ReferenceError` at import in a suite unrelated to queueing. Now constructed on first use — a queue module has no business doing construction work merely to be imported.

PR #764, opened 2026-08-02
fix(deploy): restore the deploy scripts to their cross-repo original

NateIsern:
> **Rebased onto `216b7fe5`.** #763 fixed the workflow and merged first; the workflow half of this PR is dropped as superseded. What remains is the two files #763 did not touch — and they had drifted from the copies in the other five repos.

```
.github/scripts/deploy-ecs-image.sh        +16/-1
.github/scripts/test-deploy-ecs-image.sh   +81/-40
```

## Byte-identity restored

Both files go back to blobs **`0dd66acd`** (`deploy-ecs-image.sh`) and **`9430a460`** (`test-deploy-ecs-image.sh`) — now identical in Mention, Alia, Allo, Homiio, Syra and here, mode included. That is what makes future drift checkable with `git rev-parse HEAD:<path>` instead of by reading six files. On `main` today they are `013cd6a4` / `9940af5a`.

Three things come back with them.

## 1. The harness's two ARN assertions could not fail

Measured on pristine `origin/main`, not inferred from the sibling repo:

```
$ # metrics ARN assertion pointed at a path no case uses
$ bash .github/scripts/test-deploy-ecs-image.sh
Deployment script transaction tests passed.        # exit 0
$ # same for the task-secret assertion
Deployment script transaction tests passed.        # exit 0
```

`jq -e`'s exit status was discarded. A command failing in the **middle** of a shell function consumed as `v="$(aws …)"` does not abort the run — only the function's last command reaches that assignment's exit status, and here that was a `printf`. Reproduced minimally, and it holds whether the function is exported or local, `case`-wrapped or not; a control with the failure as the last command *does* abort. `jq` had been returning 4 the whole time with nothing reading it.

Both now log a distinct token, so a mismatch lands in the `expected.log` diff and names itself.

## 2. A hyphenated-path case, and one idiom for the bracket class

`hyphenated-metrics-parameter` runs alongside the existing plain-path case, and the expected ARN is derived from the path the case feeds in rather than hardcoded in the mock.

The **jq** copy of the SSM class was never broken — Oniguruma honours `\-` — but it is the spelling that trapped the bash copy into rejecting every hyphenated path, so it is normalised to hyphen-last. One file carrying both idioms is how that bug recurs.

**Latent for oxy-api, and worth saying so:** this workflow passes no `INTERNAL_METRICS_PARAMETER` (and no `TASK_SECRET_OVERRIDES_JSON`), so the validator is never reached here today. It earns its place through byte-identity and through the next caller that does pass one — not because oxy-api is currently broken by it.

## 3. The `DEPLOY_HEAD_GUARD_SCRIPT` block #760 trimmed

Inert here — `main`'s workflow never sets `DEPLOY_SHA` (verified: no occurrence in the file) — and **verified inert rather than assumed**, by driving the *merged* workflow's own deploy step against this script with a mocked `aws`:

```
DESCRIBE service=oxy-api
REGISTER family=oxy-oxy-api container=oxy-api \
         image=…/oxy/oxy-api@sha256:7c…  secretsKept=DATABASE_URL
UPDATE   service=oxy-api taskdef=…/oxy-oxy-api:9
ECS rollout reached a healthy steady state
```

The mutable `:latest` is replaced with the digest, `DATABASE_URL` survives the re-render, no guard fires. And the guard is not vacuous — setting `DEPLOY_SHA` produces:

```
::error::DEPLOY_HEAD_GUARD_SCRIPT does not exist: .github/scripts/require-current-main.sh
```

which is the better failure than a silently skipped check.

## Verification

Mutation-tested on this rebased branch — not carried over from the pre-rebase run — each failing and naming the right thing, then restored byte-identical (`md5sum -c`):

| mutation | before | after |
|---|---|---|
| expected metrics ARN → `/bogus/NEVER` | **passed** (false green) | `-metrics:arn` / `+metrics:arn:MISMATCH` |
| expected task-secret ARN → `/bogus/NEVER` | **passed** (false green) | `-task-secret:arn` / `+task-secret:arn:MISMATCH` |
| bash class reverted to `[A-Za-z0-9_.\-/]` | n/a (no case covered it) | `Expected hyphenated-metrics-parameter to succeed` |

`Deploy Script Tests` in `ci.yml` already runs this harness on every PR, so the new case is gated here.

## Note on CI

`API Tests` is red and **that is pre-existing, not from this PR** — no file here is reachable from it. It has failed on every `CI/CD Pipeline` run on `main` since `3bec29c7` ("Merge origin/main into the Postgres port"), green on `4e58ce3a` immediately before, dying with `script "test:coverage" was terminated by signal SIGTERM`, exit 143, 8–12 minutes in. The variable duration points at a timeout or OOM rather than an assertion. Tracked separately.

A follow-up sweep will rename the harness's fixture family to something app-neutral across all six repos at once; doing it in this PR would re-diverge the file it is restoring.

### DIFF
--- a/.github/scripts/deploy-ecs-image.sh
+++ b/.github/scripts/deploy-ecs-image.sh
@@ -21,6 +21,7 @@ AWS_ACCOUNT_ID="${AWS_ACCOUNT_ID:-}"
 AWS_PARTITION="${AWS_PARTITION:-aws}"
 POST_DEPLOY_SMOKE_SCRIPT="${POST_DEPLOY_SMOKE_SCRIPT:-}"
 POST_DEPLOY_TASK_COMMAND_JSON="${POST_DEPLOY_TASK_COMMAND_JSON:-}"
+DEPLOY_HEAD_GUARD_SCRIPT="${DEPLOY_HEAD_GUARD_SCRIPT:-.github/scripts/require-current-main.sh}"
 
 if ! [[ "$MAX_WAIT_SECS" =~ ^[0-9]+$ ]] || (( MAX_WAIT_SECS < 1 )); then
   echo "::error::MAX_WAIT_SECS must be a positive integer."
@@ -36,6 +37,10 @@ if [[ "$RUN_MIGRATIONS" != "true" && "$RUN_MIGRATIONS" != "false" ]]; then
 fi
 if [[ -n "$POST_DEPLOY_SMOKE_SCRIPT" && ! -f "$POST_DEPLOY_SMOKE_SCRIPT" ]]; then
   echo "::error::POST_DEPLOY_SMOKE_SCRIPT does not exist: $POST_DEPLOY_SMOKE_SCRIPT"
+  exit 1
+fi
+if [[ -n "${DEPLOY_SHA:-}" && ! -f "$DEPLOY_HEAD_GUARD_SCRIPT" ]]; then
+  echo "::error::DEPLOY_HEAD_GUARD_SCRIPT does not exist: $DEPLOY_HEAD_GUARD_SCRIPT"
   exit 1
 fi
 if [[ -n "$POST_DEPLOY_TASK_COMMAND_JSON" ]] &&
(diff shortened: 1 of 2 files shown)


NateIsern:
Superseded by #765, which adopts this exact shape (same blobs `0dd66acd` / `9430a460`) on top of #763. Closing to avoid two PRs for one change — the duplicate assignment was mine, not a fault of this work. The false-green catch in `test-deploy-ecs-image.sh` and the hyphen-last jq normalisation both ship via #765.

PR #774, opened 2026-08-02
feat(api): make POST /auth/oauth/token speak RFC 6749 without breaking its callers

NateIsern:
## Summary

`POST /auth/oauth/token` could not be used by Matrix Authentication Service — or any conforming OAuth2/OIDC client — which blocks Allo's Matrix homeserver from using Oxy as its upstream provider. Five gaps, all fixed **on the existing route** (no `/v2`, no parallel endpoint):

1. **camelCase parameters** → RFC 6749 §4.1.3 snake_case is now accepted (`code`, `client_id`, `redirect_uri`, `client_secret`, `code_verifier`).
2. **No `grant_type`** → `authorization_code` and `refresh_token` (§6) are both implemented; an unknown grant returns `unsupported_grant_type`.
3. **No `client_secret_basic`** → HTTP Basic (§2.3.1, the MAS default) and `client_secret_post` both work. Presenting both is rejected per §2.3.
4. **Wrapped response** → RFC callers get the FLAT §5.1 body and FLAT §5.2 errors with the spec status (`invalid_client` 401 + `WWW-Authenticate`, everything else 400).
5. **No OIDC surface** → `GET|POST /auth/oauth/userinfo` and `GET /.well-known/openid-configuration` (also at the RFC 8414 path).

**Nothing changes for existing callers.** One endpoint serves two dialects and the response shape always matches the request shape. The rule is defined in exactly one place — `THE DIALECT RULE` in `packages/api/src/utils/oauthTokenRequest.ts`: an `Authorization: Basic` header or any snake_case RFC parameter selects RFC; everything else stays legacy.

### Caller audit

The only programmatic caller in the monorepo is `exchangeOAuthCode` in `packages/core/src/mixins/OxyServices.auth.ts:1205`, reached through `packages/services/src/ui/utils/oauthReturn.ts` → `runProviderColdBoot.ts` (which is how accounts/console/inbox/test-apps reach it). It sends camelCase JSON and gets byte-identical responses, asserted in the test suite. Nothing in the repo has ever sent snake_case or `grant_type`.

### Also in here

- **Refresh-token binding (§6).** §6 requires the server to ensure a refresh token was issued to the authenticated client, and an Oxy session carried no client identity. `ISession.oauthClientIds` now records every OAuth client the token endpoint minted tokens for on that session; only those may refresh it, a confidential client must authenticate, and both checks run **before** rotation so a failed caller cannot invalidate the real client's tokens. Password-login / device-flow sessions have no binding and cannot be refreshed here at all.
- `oauthCode.service.test.ts`: its `findOneAndUpdate` mock enforced single-use on the service's behalf rather than honouring the filter, so dropping the atomic `{ usedAt: null }` claim would have gone unnoticed. Fixed, and the claim is now asserted.
- Second commit is `bun run openapi:generate` output only — the artifact was stale by 62 paths, hence the size.

### Deliberately not included

No `id_token`, no JWKS. Oxy signs session tokens with an HMAC secret that has no public half, so a JWKS would be empty and an advertised signing algorithm would be a capability we cannot honour. Discovery advertises neither. **MAS must be configured with `fetch_userinfo: true`** and, if it needs a verified address, `claims_imports.email.set_email_verification` — `email_verified` is never emitted because Oxy tracks no per-address verification state.

### Security

Preserved unchanged: constant-time client-secret comparison, exact-match redirect-URI allowlist, PKCE, single-use code semantics, rate limiter. One deliberate change to the legacy path: `Cache-Control: no-store` (§5.1) is now sent on both dialects and on errors.

## Test plan

`packages/api` uses **jest**.

| | Suites | Tests |
|---|---|---|
| before | 161 | 1477 |
| after | 166 | **1569** |

All passing. `bunx tsc --noEmit` clean for `packages/api` and `packages/services`; `bun run build` (contracts → protocol → core → api) exits 0.

New coverage — 92 tests across 5 files, one per fix plus the refresh grant, discovery and userinfo:
- `src/utils/__tests__/oauthTokenRequest.test.ts` (40) — dialect rule, Basic decoding (percent-decoding, first-colon split, malformed variants), duplicate-parameter rejection, grant selection, both dialects' error shapes
- `src/routes/__tests__/oauthTokenEndpoint.test.ts` (29) — form-encoded snake_case, JSON snake_case, `client_secret_basic`, flat success body, flat error bodies with §5.2 statuses, the refresh grant and its binding, and 6 legacy-regression tests
- `src/routes/__tests__/oauthMetadata.test.ts` (10), `src/routes/__tests__/oauthUserInfo.test.ts` (8), `src/services/__tests__/session.service.oauthClientBinding.test.ts` (4)

### Mutation testing

Every rule claimed above was broken in the source and the suite re-run. **21 of 23 mutations caught**; both survivors are explained, not gaps:

| # | Mutation | Result |
|---|---|---|
| M1 | RFC parameters no longer select the RFC dialect | caught (33 fail) |
| M2 | `client_secret_basic` credentials ignored | caught (20) |
| M3 | RFC response re-wrapped in `{ data }` | caught (3) |
| M4 | `OAuthProtocolError.toJSON()` falls back to `{error,message}` | caught (5) |
| M5 | `invalid_grant` returns 401 instead of 400 | caught (5) |
| M6 | `grant_type=refresh_token` removed | caught (12) |
| M7 | refresh client binding removed | caught (2) |
| M8 | confidential client may refresh without its secret | caught (1) |
| M9 | legacy callers get the flat body | caught (1) |
| M10 | client secret never verified | caught (2) |
| M11 | discovery drops `client_secret_basic` | caught (1) |
| M12 | discovery advertises a `jwks_uri` it cannot back | caught (1) |
| M13 | userinfo wrapped in `{ data }` | caught (5) |
| M14 | userinfo `sub` becomes the mutable username | caught (4) |
| M15 | duplicate parameters accepted (§3.1) | caught (1) |
| M16 | session binding not recorded at code exchange | caught (1) |
| M16b | binding dropped on the session-reuse path | caught (1) |
| M17 | `Cache-Control: no-store` dropped | caught (3) |
| M18 | **PKCE verifier no longer compared to the stored challenge** | caught by `oauthCode.service.test.ts` (1) |
| M18b | same PKCE mutation, route tests only | **survived — expected.** Route tests mock `exchangeAuthCode`, so the real PKCE test is the one that matters, and it fires. |
| M19 | sequential `usedAt` replay guard removed | **survived — not a regression.** The atomic claim still refuses the replay; the guard is redundant defence in depth. |
| M19b | atomic `{ usedAt: null }` claim filter dropped | caught (1) — after fixing the mock that had been hiding it |
| M20 | redirect_uri binding not re-checked at exchange | caught (1) |
| M21 | exact-match redirect allowlist becomes a prefix match | caught (1) |
| M22 | `expires_in` decoupled from the real token lifetime | caught (2) |


NateIsern:
### CodeQL disposition

The CodeQL check is red on 3 alerts. Two of the four it originally reported were mine and are fixed in `d1e7b733`; the remaining three are not introduced by this change, and I have deliberately not dismissed them (that is the repo owner's call, and there is no suppression convention in this repo to follow).

**Fixed — 2× `js/polynomial-redos` (high), `utils/oauthTokenRequest.ts`.** Real and mine: the `/=+$/` padding strip used to round-trip the `Authorization: Basic` payload backtracks quadratically on attacker-controlled input at an unauthenticated endpoint. Replaced with a non-backtracking alphabet test. The round-trip was load-bearing — `Buffer.from(…, 'base64')` silently drops out-of-alphabet characters, so `aWQ6!c2VjcmV0` decodes to the same `id:secret` as the untampered string — and that case is now covered by a test that fails when the check is removed.

**Not fixed — 2× `js/insufficient-password-hash` (high).** Heuristic misfires on machine-generated, high-entropy credentials that CodeQL taints as passwords because my parser reads `client_secret`:

- `services/oauthCode.service.ts:158` — `sha256(code_verifier)` compared against the stored PKCE challenge. RFC 7636 `S256` **mandates** exactly this; a salted KDF here would break the spec and interoperability. Pre-existing code, untouched by this PR.
- `routes/auth.ts:2407` — the client-secret comparison against `ApplicationCredential.secretHash`. Pre-existing platform design (sha256 `secretHash`, documented in `AGENTS.md`); moving to a KDF is a credential-store migration across every credential type, not something to smuggle into an OAuth conformance PR. The logic is byte-identical to before, only relocated into `authenticateOAuthClient`.

`main` already carries an open alert of this same rule (`packages/core/src/mixins/OxyServices.auth.ts:469`).

**Not fixed — `js/missing-token-validation` (high), `server.ts:111`.** The global cookie-parser/CSRF finding, unrelated to this change. `main` carries the identical alert at `server.ts:141`; the line differs only because of where this branch forked. CodeQL's own summary notes: _"Alerts not introduced by this pull request might have been detected because the code changes were too large."_

NateIsern:
Cerrado sin fusionar: este trabajo ya está en `main` desde #756 (`feat(auth)!: el endpoint de token habla RFC 6749, y aparece userinfo`), fusionado el 2026-08-02 03:43, más el seguimiento `64386049`.

La rama se cortó desde `feat/canonical-text-normalizers` sin hacer `fetch`, así que el análisis que la motivó se hizo sobre un árbol anterior a #756 y concluyó que faltaba una superficie que ya existía. Verificado ahora contra `origin/main`: exige `grant_type=authorization_code` y responde `unsupported_grant_type` al resto, acepta `client_secret_basic` y `client_secret_post` (`utils/oauthClientAuth.ts`), devuelve cuerpo plano §5.1 con `Cache-Control: no-store` y errores §5.2 (`utils/oauthResponse.ts`), y sirve `GET|POST /auth/oauth/userinfo` con `sub` = `_id`.

Donde este PR difiere, contradice decisiones ya fusionadas y marcadas como ruptura: #756 rechaza `application/json` a propósito y declara que Oxy no emite refresh tokens a clientes OAuth, porque el par `deviceId`+`deviceSecret` los sustituye. Rebasar habría reabierto ambas por la puerta de atrás de una resolución de conflictos.

Las dos alertas de CodeQL desaparecen con la rama: nacían del taint desde `utils/oauthTokenRequest.ts`, que solo existe aquí.

Lo que queda de valor está registrado en oxy-infra#21: MAS va configurado con `discovery_mode: disabled` y endpoints explícitos, así que el documento de discovery no hace falta.

PR #777, opened 2026-08-02
chore(release): contracts 0.22.0, core 17.1.0

NateIsern:
Releases the `channel` account kind (#776) so Mention can consume it.

Contracts moves first and publishes first: it now owns the account-kind vocabulary that the Drizzle schema, the Mongoose model, the SDK union and the create-account validator all derive from, and `core` depends on it via `workspace:^` — which `bun pm pack` rewrites to a caret on the **local** version. Publishing core against a contracts version not yet on the registry ships a manifest nobody can install.

Lockfile committed in the same commit as the manifests, per the house rule.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -692,7 +692,7 @@
     },
     "packages/services": {
       "name": "@oxyhq/services",
-      "version": "26.0.2",
+      "version": "26.1.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


NateIsern:
Superseded — the version bumps ride with the blockers in the same PR so Oxy publishes once.

PR #855, opened 2026-08-07
fix(core)!: refuse user tokens that are not bound to a validated session

NateIsern:
## Summary

`oxy.auth()` decoded the bearer JWT with `jwtDecode` (no signature verification) and then trusted what it found:

1. **A user token with no `sessionId` was accepted on its claims alone.** No network call happened; `req.userId` came straight off the decoded `userId`. Anyone could authenticate as any user on any Oxy backend with a hand-rolled JWT carrying a `userId`, a future `exp`, and an invented signature.
2. **A user token *with* a `sessionId` had its session validated, but `req.userId` still came from the token.** `GET /session/validate/:sessionId` is unauthenticated and does not bind the bearer, so anyone holding a live session id could pair it with a forged `userId` and be trusted as that user.

A user token must now carry a `sessionId`, that session is validated server-side, and the identity is read **off the validated session** — `SESSION_REQUIRED`, `SESSION_USER_MISMATCH`, `INVALID_SESSION`. Under `optional: true` all three give anonymous rather than the claimed user. `authSocket()` already enforced this; the HTTP middleware now matches it.

Nothing legitimate breaks: every user access token the Oxy API issues carries a `sessionId`. The only session-less user JWTs it mints are the 2FA challenge and account-recovery tokens, both pre-authentication and both previously accepted as a full logged-in identity.

Also makes `createOxyRateLimit`'s session resolution actually idempotent (it uses `createOptionalOxyAuth` now, so it no longer clobbers an identity a preceding middleware resolved), which lets an app running a second auth scheme mount it ahead of the limiter and get a per-user bucket.

`@oxyhq/core` 20.0.0 → 21.0.0.

## Test plan

- New `packages/core/src/mixins/__tests__/userTokenAuth.test.ts`: 17 cases — forged signature, empty signature, `alg: none`, `id` claim instead of `userId`, the real 2FA challenge and recovery token shapes, `loadUser`, `optional`, `onError`; and on the session path a forged `userId` against a live session, a session with no user, `_id`-shaped identities, expiry, and the legitimate paths still working. **12 of them fail on `main` and pass here.**
- `packages/core/src/server/__tests__/rateLimit.test.ts`: rewrote the case whose premise was "`oxy.auth()` can produce an unverified non-session identity" (that path no longer exists) and added idempotency + service-token bucket coverage.
- `bun run core:build`, `tsc --noEmit`, and the full core suite: **1572 tests, 113 suites, all passing.**

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -424,7 +424,7 @@
     },
     "packages/core": {
       "name": "@oxyhq/core",
-      "version": "20.0.0",
+      "version": "21.0.0",
       "dependencies": {
         "@noble/ciphers": "^1.3.0",
         "@noble/hashes": "^1.8.0",
@@ -768,7 +768,7 @@
       "peerDependencies": {
         "@expo/vector-icons": "^15.0.3",
         "@oxyhq/bloom": ">=0.59.0",
-        "@oxyhq/core": "^20.0.0",
+        "@oxyhq/core": "^21.0.0",
         "@react-native-async-storage/async-storage": "^2.0.0",
         "@react-native-community/netinfo": ">=11.4.1",
         "@tanstack/query-async-storage-persister": "catalog:",
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,74 @@
 # Changelog — `@oxyhq/core`
 
+## 21.0.0
+
+### Security: `oxy.auth()` accepted forged user tokens
+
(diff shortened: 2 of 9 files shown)


NateIsern:
Superseded by #856, now merged. Same session-binding fix, and #856 additionally closes the claim-swap variant on an attacker's own session (GET /session/validate/:id is unauthenticated and unbound), carries mutation tests for both guards, and preserves the newer service-token environment isolation and rate-limiter IP-hashing that this branch's base predated. Closing to prevent a double merge.

PR #881, opened 2026-08-08
feat(create-oxy-app): scaffold on Postgres/drizzle instead of MongoDB

### LINKED ISSUE #871: create-oxy-app still scaffolds new Oxy apps on MongoDB
Found during the Mongo-trace sweep (2026-08-08): packages/create-oxy-app generates a mongoose connector (templates/backend/.../src/config/database.ts) and MONGODB_URI in the generated .env.example — 6 files. Every new Oxy app is currently born on Mongo while the ecosystem's backends run PostgreSQL (mention, oxy-api, allo-matrix, mercaria). Whether new apps should scaffold on Postgres (drizzle + @oxyhq/db, the Mercaria port as the reference shape) is a PRODUCT decision — flagging rather than deciding.

NateIsern:
Closes #871

`create-oxy-app` generated every new Oxy app with a Mongoose connector, so each one started life on the store the ecosystem is migrating off. The backend template now ships the Oxy Postgres substrate — `@oxyhq/db` + `drizzle-orm` + `postgres.js`.

## What the generated backend gets

- `src/db/postgres.ts` — `connectPostgres` / `getDb` / `checkPostgresHealth` / `closePostgres`, built through `@oxyhq/db`'s `createDatabase` so the runtime handle and drizzle-kit read the *same* `DATABASE_CASING`.
- `src/db/migrate.ts` — drizzle-orm's own migrator (never `drizzle-kit migrate`, which is a devDependency and cannot reach production), carrying the deploy-phase and `--target-database` guards.
- `src/db/schema/` — barrel plus one example `notes` table demonstrating `generatedId()`, `timestamptz`, a CHECK derived from a `const` tuple, and an id column deliberately carrying no FK because Oxy owns identity.
- `drizzle/0000_init.sql` + snapshot, `drizzle.config.ts`, and a root `docker-compose.postgres.yml`.
- `/ready` now does a real round trip instead of reading a cached "we connected once" flag.

## Three decisions worth reviewing

**The migrations folder is found by searching upward, not computed.** `tsconfig` compiles with `rootDir: "./"`, so `src/db` sits two levels below the package root and `dist/src/db` sits three. A fixed `path.join(__dirname,'..','..','drizzle')` therefore works from source and resolves to `dist/drizzle` in the image — green in dev, broken in the container. Mutation-tested in both positions: from source the naive path reports `journal entries=1`; from `dist` it dies on `.../dist/drizzle/meta/_journal`.

**The server connects but never migrates on boot.** Every task in a scaled service would run it at once, drizzle's migrator takes no lock, and it would defeat the pre/post phase split entirely. Migrations are their own one-shot step, and the runtime image now copies `drizzle/` next to the compiled migrator.

**`drizzle.config.ts` is excluded from the emitting build** — it imports `drizzle-kit`, which the runtime image strips, so compiling it placed a module in `dist/` that throws if anything requires it.

## Verification

End to end from the **packed tarball**, not the repo tree: scaffold → `bun install` → typecheck → build → migrate a real Postgres 17 **from zero via the compiled `dist/src/db/migrate.js`** → re-run as a no-op → confirm a wrong `--target-database` is refused. Deploy phases were exercised both ways (a `post` migration deferred by `--phase=pre`, then applied by `--phase=post`), an unmarked migration is refused before any DDL, and `/ready` was confirmed to fall to 503 with Postgres stopped while `/health` stays 200.

`db:generate` on a fresh scaffold reports **no drift**, so no generated app starts with a phantom migration its first generate invents.

## CI

`scaffold-smoke` gains a Postgres service container and asserts the rendered Postgres layer, the pinned dependency set, migration from zero, idempotent re-run, wrong-target refusal, and zero drift. Six new unit tests cover the template pieces; every gate here was mutation-tested red before being trusted, and each carries a vacuity floor so a broken traversal cannot pass by examining nothing.


NateIsern:
Closing as a duplicate of #872, which landed on `main` while this was in flight and is a superset of it — same template layout, same `migrationsFolder` upward search, same deploy-phase markers, plus a more thorough unit suite (marker near-miss cases, compose/DATABASE_URL consistency, and a deliberate non-5432 port).

One thing here is not in #872 and I'll send it as a focused follow-up rather than as this branch: the generated `packages/backend/tsconfig.json` still compiles `drizzle.config.ts` into `dist/`. That config imports `drizzle-kit`, a devDependency the runtime image strips, so the built image carries a module that throws if anything ever requires it.

PR #898, opened 2026-08-10
fix(app-preset): require OTA code signing by default

NateIsern:
### Motivation
- Prevent production OTA-enabled binaries from being built in a fail-open state that accepts unsigned manifests.  
- The previous default (`codeSigning: 'auto'`) wired `updates.url` even when the bundled certificate was absent, leaving clients unable to request signed manifests and enabling unsigned JS delivery if the update path is tampered with.  
- Make secure behavior the default while still allowing explicit development-only opt-outs.

### Description
- Change the `withOxyUpdates` config plugin default from `codeSigning: 'auto'` to `codeSigning: 'require'` in `packages/app-preset/plugin/withOxyUpdates.js` so builds fail when the certificate is missing.  
- Update tests in `packages/app-preset/__tests__/withOxyUpdates.test.js` to assert the new secure default and add a helper `unsignedOptions` for tests that intentionally exercise unsigned behavior.  
- Clarify the preset and certs documentation in `packages/app-preset/README.md` and `packages/app-preset/certs/README.md` to state the secure default and describe explicit development modes (`auto` for warnings, `false` to opt out).  
- Preserve explicit `codeSigning: 'auto'` and `codeSigning: false` modes for development/testing scenarios.

### Testing
- Attempted `bun run --cwd packages/app-preset test` initially failed due to a missing peer `expo/config-plugins` in the environment, which is expected in a minimal environment.  
- Ran the package tests with a temporary minimal `expo/config-plugins` stub via `NODE_PATH` and `bun run --cwd packages/app-preset test`, and all tests in `packages/app-preset` passed (11/11).  
- Ran `git diff --check` to validate whitespace/patch hygiene with no issues reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a73febf1b6c83238cd3dd5ea802d731)

### DIFF
--- a/packages/app-preset/README.md
+++ b/packages/app-preset/README.md
@@ -103,7 +103,7 @@ prebuild applies its config plugin automatically.
 | `clientId` | required | The registered `ApplicationCredential` publicKey |
 | `apiOrigin` | `https://api.oxy.so` | Origin serving `/updates/v1` |
 | `channel` | `production` | Release channel this binary polls |
-| `codeSigning` | `auto` | `require` to fail the build when the certificate is missing; `false` to skip signing |
+| `codeSigning` | `require` | Fails when the certificate is missing; `auto` warns and `false` skips signing (development only) |
 | `certificatePath` | bundled certificate | Only for a key-rotation overlap |
 | `runtimeVersionPolicy` | `appVersion` | `false` leaves whatever the app declared |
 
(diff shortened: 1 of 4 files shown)


NateIsern:
Reverted on `main` in `e3bde5fd1` — the default flip cannot land until the certificate exists.

I merged this on a full-green run (32 checks). That green was wrong, and the mechanism is worth recording: `deploy-accounts` runs only on `push` to `main` and is path-filtered, so the job this breaks **did not exist for the PR**. It went red on the merge commit and was `success` on the commit before it.

The failure:

```
Error: [@oxyhq/app-preset/plugin/withOxyUpdates] code signing is required but the
Oxy Updates certificate is missing at
packages/app-preset/certs/oxy-updates-code-signing.pem. Generate the ecosystem
keypair with oxy-api's `bun scripts/generate-updates-code-signing.ts`, commit the
certificate into @oxyhq/app-preset, and set the private key as
UPDATES_CODE_SIGNING_PRIVATE_KEY on oxy-api.
```

`packages/app-preset/certs/` contains only `README.md` — the ecosystem keypair was never generated. So `require` is not a stricter default for unconfigured apps, it is a build failure for **every** app using the preset, `accounts` included.

The revert touches the default only. The require/auto branch, the error message and the README you added all stay, so this becomes a one-line change again once:

1. the keypair is generated with oxy-api's `generate-updates-code-signing.ts`,
2. `oxy-updates-code-signing.pem` is committed into `@oxyhq/app-preset`, and
3. `UPDATES_CODE_SIGNING_PRIVATE_KEY` is set on oxy-api.

Reopening this so the change is not lost — it should land as the last step of that sequence, not the first.

Worth fixing separately: a preset default that only `deploy-accounts` can catch means any future change to this plugin merges unverified. Building one preset consumer in `CI/CD Pipeline` would have caught it on the PR.

PR #899, opened 2026-08-10
Require application administrators for follow registry writes

NateIsern:
### Motivation
- The follow-registry routes treated a user-delegated `follow-targets:register` grant as sufficient authority to mutate application-global registry state, letting ordinary OAuth grantees claim namespaces and change kinds for an application they do not administer.
- The intent is to keep `follow-targets:register` as a necessary consent for follow-related operations while preventing user-scoped grants from becoming application-administration authority.

### Description
- Harden `requireRegistrar` in `packages/api/src/routes/followRegistry.v2.routes.ts` to, after verifying `follow-targets:register`, load the target application's `ownerAccountId` and require the caller to have `owner` or `admin` effective access via `accountService.resolveEffectiveAccess` before permitting registry writes.
- Add imports and a small DB lookup (`getDb()` + `applications.ownerAccountId`) to determine the owning account for the application referenced by the capability.
- Update the test helper `signIn` in `packages/api/src/routes/__tests__/followsV2.test.ts` to support creating an application not owned by the token-holder and add an HTTP regression test asserting a fully scoped but non-administrative user receives `403` when attempting to claim a namespace.

### Testing
- Ran `git diff --check` with no whitespace or trivial patch errors reported. 
- Added an automated Jest regression test in `packages/api/src/routes/__tests__/followsV2.test.ts`, but running `bun run test -- --runInBand src/routes/__tests__/followsV2.test.ts` could not be completed in this environment because repository dependencies (Jest/Bun toolchain) are not installed and `jest: command not found` was returned. The new test is present and ready to run in CI where dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71652d05c483239b4c4c5dd354a88e)

### DIFF
--- a/packages/api/src/routes/__tests__/followsV2.test.ts
+++ b/packages/api/src/routes/__tests__/followsV2.test.ts
@@ -93,12 +93,20 @@ async function request(
 }
 
 /** Mint a user, an application, an authorization and a grant with these scopes. */
-async function signIn(scopes: string[]): Promise<{ userId: string; applicationId: string }> {
+async function signIn(
+  scopes: string[],
+  options: { ownedByUser?: boolean } = {}
+): Promise<{ userId: string; applicationId: string }> {
   const db = getDb();
   const [user] = await db.insert(users).values({}).returning({ id: users.id });
+  let ownerId = user.id;
+  if (options.ownedByUser === false) {
+    const [owner] = await db.insert(users).values({}).returning({ id: users.id });
+    ownerId = owner.id;
+  }
   const [app] = await db
     .insert(applications)
-    .values({ name: unique('App '), status: 'active', ownerAccountId: user.id })
+    .values({ name: unique('App '), status: 'active', ownerAccountId: ownerId })
     .returning({ id: applications.id });
 
   const sessionId = unique('session-');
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (branch is 247 commits behind). **Premise partly live, but the patch needs a product decision before it can land.**

Still true on `main`: `requireRegistrar` checks only that the session came from an application authorization with the `register` follow scope. It does not check the caller's role on the application's owning account.

Not the same as when this was written: `main` has since added per-application ownership *at the registry layer* — `namespace_taken`, `namespace_not_owned`, `kind_not_owned` (see `feat(follows): namespaces have owners, so one app cannot define another's kinds`). So the cross-application escalation is already closed. What remains open is narrower: *any user* of application X, holding an X-scoped grant, can claim namespaces and register kinds **on X's behalf**.

The blocker is that the patch requires `owner`/`admin` on the application's owning account, and this route is user-authenticated (`authMiddleware` + `resolveFollowCapability(userId, sessionId)`). If any app registers its namespaces from a normal user session — a boot-time or first-use registration — this turns that into a 403 for every non-admin user. That is the "breaks the legitimate path" shape, and I can't rule it out from the repo alone.

Two questions decide it:

1. Is registry write meant to be an **application-operator** action (then this patch is right, and registration should move to a service credential, not a user session), or an **app-delegated** one (then the `register` scope *is* the intended delegation and this patch is wrong)?
2. If the former: does anything currently register via a user session? Those call sites need moving first, or this is a breaking change.

Leaving open pending that call.

PR #900, opened 2026-08-10
fix(api): prevent WebFinger actor relabelling

NateIsern:
### Motivation

- Prevent a high-impact integrity vulnerability where an attacker-controlled WebFinger resource could point its self-link at an already-cached actor URI and cause the upsert path to overwrite that actor's stored username and federation domain.

### Description

- Treat an existing `federationActorUri` as the stable identity key and immediately return the cached federated row instead of falling through to the fetch/upsert path in `packages/api/src/services/federation.service.ts`.
- Remove the previous conditional that allowed non-bridge actor rows to be re-attributed via WebFinger and preserve intentionally relabelled bridge identities by returning the cached row unconditionally when present.
- Replace the prior upsert-focused test with a regression test in `packages/api/src/services/__tests__/federation.service.test.ts` that simulates a malicious WebFinger alias targeting an existing actor and asserts the cached victim remains unchanged and the attacker profile is not fetched.
- Files changed: `packages/api/src/services/federation.service.ts` and `packages/api/src/services/__tests__/federation.service.test.ts`.

### Testing

- Ran `git diff --check`, which reported no whitespace/patch problems and completed successfully.
- Attempted `bun run test --runTestsByPath src/services/__tests__/federation.service.test.ts --runInBand`, but the test run failed in this environment because `jest` is not available (repository dependencies are not installed), so the unit test could not be executed here.
- The change is covered by the new regression test added to `packages/api/src/services/__tests__/federation.service.test.ts`, which should be executed in CI or a developer environment with dependencies installed to validate behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a713ce545308323834df874bc85d0d5)

### DIFF
--- a/packages/api/src/services/__tests__/federation.service.test.ts
+++ b/packages/api/src/services/__tests__/federation.service.test.ts
@@ -538,35 +538,34 @@ describe('FederationService.resolveAndUpsert (fast + eventually-fresh)', () => {
     expect(result?.avatar).toBe('new-file-id');
   });
 
-  it('upserts onto the EXISTING row when the actor URI is already known', async () => {
-    const fx = nextFixture();
-    // Seeded under a DIFFERENT username/domain, so the cached-user lookup
-    // misses and the insert path is taken — it must then collide on
-    // `users_federation_actor_uri_key` and update rather than duplicate.
-    const userId = await seedFederatedUser(fx, STALE_AGE_MS, {
-      username: `stale-${actorCounter}@elsewhere.example`,
-      federationDomain: 'elsewhere.example',
-    });
+  it('does not let a WebFinger alias relabel an existing actor URI', async () => {
+    const victim = nextFixture('victim.example');
+    const attackerHandle = `attacker${actorCounter}@evil.example`;
+    const userId = await seedFederatedUser(victim, STALE_AGE_MS);
 
-    webfingerSpy.mockResolvedValue({ actorUri: fx.actorUri, subjectAcct: fx.handle });
+    // The attacker's server points its self link at an already-cached victim.
+    webfingerSpy.mockResolvedValue({
+      actorUri: victim.actorUri,
+      subjectAcct: attackerHandle,
+    });
     actorSpy.mockResolvedValue({
-      actorUri: fx.actorUri,
-      domain: fx.domain,
-      username: fx.handle,
-      displayName: 'Alice Moved',
+      actorUri: victim.actorUri,
+      domain: 'evil.example',
+      username: attackerHandle,
+      displayName: 'Attacker',
       avatarUrl: undefined,
       bio: undefined,
     });
 
-    const result = await federationService.resolveAndUpsert(fx.handle);
+    const result = await federationService.resolveAndUpsert(attackerHandle);
 
-    // One row, updated — never a second row for the same actor.
     expect(result?._id).toBe(userId);
-    expect(await storedByActorUri(fx.actorUri)).toMatchObject({
+    expect(result?.username).toBe(victim.handle);
+    expect(actorSpy).not.toHaveBeenCalled();
+    expect(await storedByActorUri(victim.actorUri)).toMatchObject({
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (branch is 247 commits behind). **The vulnerability is live and I traced it end to end. The patch works but costs a documented repair path — read the alternative below before merging.**

## The attack, on today's `main`

`resolveFederatedUser` → `fetchActorProfile(actorUri, verifiedAcct)`, and in `fetchActorProfile`:

```ts
const acct = hintedAcct || actorWebfinger || `${username.toLowerCase()}@${actorHost}`;
const domain = domainFromHandle(acct) || actorHost;
```

`hintedAcct` is `verifiedAccountForResolution(cleaned, webfinger)`, and that function verifies **only when the WebFinger subject differs from the requested handle**:

```ts
if (!requested || !subject || subject === requested) {
  return requested || requestedAcct.toLowerCase();
}
```

So an attacker who controls `evil.com`:

1. Serves WebFinger for `victim@evil.com` with `subject: "acct:victim@evil.com"` and a self link pointing at `https://mastodon.social/users/victim` — a real, existing actor URI.
2. `subject === requested`, so `verifiedAccountForResolution` returns `victim@evil.com` **unverified**.
3. `existingByActorUri` finds the victim's row. `bridgeVouchesForNetwork('mastodon.social', <stored domain>)` is false, so it falls through — exactly as the comment intends for a stale row.
4. `fetchActorProfile` fetches the genuine mastodon.social actor, but `acct` is the *hint*, so `username: 'victim@evil.com'`, `domain: 'evil.com'`.
5. The upsert is `onConflictDoUpdate({ target: users.federationActorUri, set: setFields })` — keyed on the actor URI — so it **rewrites the victim's username and federation domain in place**.

The victim's federated identity is relabelled to the attacker's domain, without the attacker ever controlling the actor.

## Why I did not merge this as-is

The patch returns cached unconditionally whenever a row exists for the actor URI. That closes the hole, but it also deletes the case the current comment calls out explicitly:

> A stale row on a NON-bridge actor (wrong domain stored beside a canonical actor URI) must fall through to the upsert.

After this change a row with a wrong stored domain can never self-correct through this path.

## The narrower fix

The defect is not the fall-through — it is that an **unverified hint decides the stored handle**. Gate the hint instead, in `fetchActorProfile`:

```ts
const hintHost = hintedAcct ? domainFromHandle(hintedAcct) : null;
const hintTrusted =
  hintHost !== null
  && (isSameFederationHost(hintHost, actorHost) || bridgeVouchesForNetwork(actorHost, hintHost));
const acct = (hintTrusted ? hintedAcct : null) || actorWebfinger || `${username.toLowerCase()}@${actorHost}`;
```

That blocks the relabelling (the hint's domain must be the actor's host, or a bridge must vouch for it), keeps relabelled bridge identities working, and leaves the stale-row repair intact.

Closing the `subject === requested` shortcut in `verifiedAccountForResolution` is worth considering separately, but it is not sufficient on its own — the fallback `return requested` at the end of that function reaches the same place.

Leaving open for your call on which of the two shapes to land. Either way this should not sit — the current behaviour is exploitable by anyone who can host a WebFinger endpoint.

NateIsern:
Landed the narrower fix as #1189, which supersedes this diff.

It gates the **hint** in `fetchActorProfile` instead of refusing to fall through, so the relabelling is closed while the stale-row repair this PR would have removed stays working.

Verified with a positive control — reverting only the fix and keeping the tests reproduces the vulnerability exactly as traced here:

```
● refuses a hint from a host the actor does not vouch for
  Expected: "victim@mastodon.social"   Received: "victim@evil.example"
Tests: 2 failed, 32 passed, 34 total
```

Closing this one in favour of #1189 — say the word if you would rather keep it open as the tracking issue for the `verifiedAccountForResolution` shortcut, which #1189 deliberately leaves alone.

PR #902, opened 2026-08-10
fix(auth): rotate device secret after token mint

NateIsern:
### Motivation

- A recent change stopped rotating the device credential in the public `POST /session/device/token` mint, leaving a captured `deviceSecret` usable indefinitely and enabling replayed mints that prolong session lifetime.

### Description

- Re-enabled in-place rotation in the mint handler by calling `deviceSessionService.issueDeviceSecret(deviceId)` after a successful token resolution and returning the fresh `nextDeviceSecret` in the response.
- The handler now fails closed with `500` if rotation cannot be completed for a live mint, while unchanged behavior remains for unsuccessful mints (they keep the existing credential).
- Updated unit tests in `packages/api/src/routes/__tests__/sessionDevice.test.ts` to assert the new rotation behavior, that the previous secret moves to the grace slot, and that the superseded secret is rejected after the grace window.
- Updated documentation in `docs/SESSION-ARCHITECTURE.md` to describe rotation-in-use and the concurrency grace window semantics (60s) so the transport contract matches implementation.

### Testing

- Ran repository checks via `git diff --check` which reported no whitespace errors and the working tree is consistent with the change set.
- Attempted to install dependencies with `bun install --frozen-lockfile`, but the environment could not fetch registry packages (HTTP 403), preventing a full test run.
- Attempted to run the package tests with `bun run --filter @oxyhq/api test`, but `jest` was unavailable because dependencies were not installed, so unit suites could not be executed here.
- The modified test cases were updated to reflect the intended secure behavior (rotation + grace semantics) and are ready to run in CI or a local environment with dependencies available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71192854c88323b60feebc94e23d8e)

### DIFF
--- a/docs/SESSION-ARCHITECTURE.md
+++ b/docs/SESSION-ARCHITECTURE.md
@@ -50,7 +50,7 @@ from the request body.
 
 | Method | Route | Body | Behavior |
 |--------|-------|------|----------|
-| POST | `/session/device/token` | `{ deviceId, deviceSecret }` | **The zero-cookie mint** — PUBLIC (no bearer, no cookies): possession of the secret is the device-ownership proof. Verifies `sha256(deviceSecret)` (constant-time) against the device's `secretHash`, mints a short access token for the active account, and returns the proven secret unchanged as `nextDeviceSecret`. Keeping the credential stable lets multiple official apps/origins sharing one device refresh concurrently. Per-device lockout + rate limit blunt online guessing. |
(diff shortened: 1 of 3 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (branch is 247 commits behind). **Rejecting: this is a straight revert of a deliberate fix.**

`main` carries `96676f6f4 fix(auth): keep shared device credentials stable during mint`, and the comment this PR rewrites states the reason it exists:

> Keep the proven credential stable. Rotating here makes separate first-party origins race over one DeviceSession secret and causes otherwise healthy sessions to disappear after the grace window.

The `deviceSecret` is a per-DEVICE credential shared by several first-party origins on the same device. Rotating it on every mint means whichever origin refreshes second presents a secret that is now the `prevSecretHash`, and once the grace window lapses those sessions are gone — which is the user-visible spurious-logout failure that commit was landed to stop.

Reverting it to get credential rotation would reintroduce that. If rotation is still wanted, it needs a design that does not have several origins sharing one rotating secret — per-origin credentials, or a rotation the other origins can observe — not a flip of this default.

Closing.

PR #903, opened 2026-08-10
fix(api): retain shared actors during domain purge

NateIsern:
### Motivation

- A vulnerability allowed a federation-enabled app to hard-delete globally shared federated actor rows when no other app-tagged files existed, which can destroy another tenant's social graph when that tenant referenced the actor but did not upload files. The change closes this authorization/retention gap by treating file attribution as insufficient proof of exclusive ownership.

### Description

- Change purge semantics to only remove files owned by the caller and to always retain/archive the globally keyed federated actor row and its platform avatars/graph instead of hard-deleting them. (`packages/api/src/services/federation/blockedDomainPurge.service.ts`)
- Add a `retentionReason` discriminator to `RetainedActor` so responses distinguish known other-app file references from unknown potential references. (`blockedDomainPurge.service.ts`) 
- Stop reusing the actor hard-delete path from the purge; actor teardown is no longer invoked by domain purge (actor rows are archived instead). (`blockedDomainPurge.service.ts`) 
- Update the route documentation to explain the conservative authorization boundary (caller-owned files only; shared actors retained). (`packages/api/src/routes/federation.ts`) 
- Update tests to reflect the conservative behavior and add a regression test ensuring an actor + local follow survive a real purge even when no other application-tagged file exists, while the caller's own file is deleted; adjust expectations for avatars and pagination/resumption. (`packages/api/src/routes/__tests__/federationDomainPurge.test.ts`)

### Testing

- Ran repository checks and committed the change as `fix(api): retain shared actors during domain purge` (commit present in the working tree). (`git diff --check`, `git commit`) — succeeded.
- Attempted dependency install with `bun install --frozen-lockfile`, but registry artifact fetches returned HTTP 403 in this environment, so install failed and test runners could not be executed end-to-end (environment limitation). — failed to install dependencies.
- Attempted to run the single-test invocation via the test script, but `jest`/test tooling was unavailable due to the failed install, so tests could not be executed here. — could not run Jest.
- All changes are covered by updated/added tests under `packages/api/src/routes/__tests__/federationDomainPurge.test.ts`; CI should run the full suite (including the updated Jest tests) after dependencies are available and will verify behavior in a real Postgres-backed environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7121fff4748323962c7b963eee8fd4)

### DIFF
--- a/packages/api/src/routes/__tests__/federationDomainPurge.test.ts
+++ b/packages/api/src/routes/__tests__/federationDomainPurge.test.ts
@@ -265,6 +265,15 @@ async function follow(followerId: string, followedId: string): Promise<void> {
   await getDb().insert(userFollows).values({ followerId, followedId });
 }
 
+async function followExists(followerId: string, followedId: string): Promise<boolean> {
+  const [row] = await getDb()
+    .select({ id: userFollows.id })
+    .from(userFollows)
+    .where(and(eq(userFollows.followerId, followerId), eq(userFollows.followedId, followedId)))
+    .limit(1);
+  return row !== undefined;
+}
+
 async function userExists(id: string): Promise<boolean> {
   const [row] = await getDb().select({ id: users.id }).from(users).where(eq(users.id, id)).limit(1);
   return row !== undefined;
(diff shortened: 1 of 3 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f`. **Fail-safe and defensible, but it is a product decision, not a bug fix — so I am not merging it unilaterally.**

The argument is sound. `blockedDomainPurge` decides "nothing else references this actor" from file attribution alone, and the module's own comment concedes that this is a fact about the current deployment rather than about the data model:

> Today exactly one app federates … That is a fact about the current DEPLOYMENT, not about the data model, and it stops being true silently the day a second app federates.

An app with no *tagged files* on an actor may still reference it, so file absence is not proof, and the consequence of being wrong is an irreversible cross-tenant delete.

What this changes, and why it needs your call rather than mine:

- Actor rows are now **always** retained and archived. The endpoint stops removing `type:'federated'` rows and their social-graph edges entirely, so a blocked instance leaves archived actors and edges behind indefinitely.
- Platform-fetched avatars (the `unattributed` files) are no longer deleted either, so `avatarsDeleted` becomes dead in practice and the storage the purge was partly meant to reclaim stays.
- `RetainedActor` grows a `retentionReason`, which is a response-shape change for any caller parsing the plan.

That is a real narrowing of a documented endpoint's contract, in exchange for removing a delete that cannot be undone. I would take that trade, but it should be your decision and it probably wants a note in the federation docs that a domain purge no longer removes actors.

Leaving open. If you want it, it rebases cleanly onto `main` — say so and I will land it.

PR #904, opened 2026-08-10
fix(api): block unauthenticated lazy MP4 transcoding

NateIsern:
### Motivation
- Prevent unauthenticated public media requests from triggering on-demand FFmpeg MP4 transcoding (360p/720p/1080p) via `ensureVariant`, which allowed unbounded CPU and memory use and concurrent ffmpeg spawns from attacker-controlled `variant` query values.

### Description
- Stop lazy-generation of MP4 bitrate renditions in `AssetService.ensureVariant` by throwing an explicit error for MP4 rendition names instead of calling `ensureVideoMp4Rendition`, preserving serving of already-ready variants and poster/thumb/image flows.
- Update the routing unit test `src/services/__tests__/assetService.ensureVariantVideo.test.ts` to assert that a missing MP4 rendition is rejected and that `ensureVideoMp4Rendition` is not invoked while poster/thumb behavior remains covered.
- This change keeps upload-time rendition generation and existing variant storage/serving logic intact while removing the expensive work from public request paths.

### Testing
- Ran `git diff --check` to validate the working tree changes and found no whitespace/patch issues (check passed). 
- Attempted to run the targeted unit test via `bun run test --runInBand src/services/__tests__/assetService.ensureVariantVideo.test.ts`, but the test runner failed to execute in this environment because `jest` is not available (`exit code 127`), so the test could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71217fef748323a3a16863dd064677)

### DIFF
--- a/packages/api/src/services/__tests__/assetService.ensureVariantVideo.test.ts
+++ b/packages/api/src/services/__tests__/assetService.ensureVariantVideo.test.ts
@@ -1,9 +1,8 @@
 /**
- * AssetService.ensureVariant — video mp4 rendition routing (#759).
+ * AssetService.ensureVariant — safe video mp4 rendition routing.
  *
- * Upload-time generation can leave a video with poster/hls_master but no
- * `360p`/`720p`/`1080p`. `ensureVariant` must lazily regenerate those on
- * request instead of throwing "not supported for mime video/*".
+ * Public media routes reach ensureVariant, so missing MP4 renditions must not
+ * start expensive FFmpeg work on the request path.
  */
 
 import { AssetService } from '../assetService';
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (branch is 247 commits behind). **The finding is live; the patch is not the fix.**

The finding stands. `GET /assets/:id/stream` uses `optionalAuthMiddleware`, so for a public file an unauthenticated caller reaches `assetService.ensureVariant` → `variantService.ensureVideoMp4Rendition` → FFmpeg. There is no in-flight dedup anywhere in `variantService.ts` (`git grep -n "inFlight\|Map<string, Promise"` → nothing), so N concurrent requests for the same missing rendition start N transcodes.

But the patch throws unconditionally, which reverts `#759`. `ensureVideoMp4Rendition` on `main` is deliberate self-healing, documented as such:

> Upload-time `generateVideoVariants` swallows per-rendition failures (`resolve(null)`), so a file can have `poster`/`hls_master` but no `360p`/`720p`/`1080p`. Lazy generation here lets those files self-heal on first request (#759).

Merging this would permanently break playback for exactly the files that feature exists for — the route's `catch` turns the throw into a 404 or a placeholder.

What the fix should be instead, keeping the self-heal:

1. An in-flight promise map keyed on `${file.id}:${variantType}`, so concurrent requests join one transcode.
2. A global concurrency cap on FFmpeg jobs, rejecting (not queueing unboundedly) past it.
3. Optionally require an authenticated viewer or a scoped `?mt=` media token *for generation only*, still serving an already-generated rendition anonymously.

The existing bounds — `getUsableReadyVariant` short-circuit, the `videoVariants` allowlist, and the source-resolution guard — already make this once-per-asset-per-rendition rather than unbounded per request, so this is a rate/concurrency problem, not an authorization one.

Leaving open so the finding is not lost, but the current diff should not be merged as-is.

PR #905, opened 2026-08-10
Enforce exact federation authority host boundaries

NateIsern:
### Motivation

- Fix a security regression where `www.`-prefixed sibling hosts were treated as equivalent to their apex host for federation authorization, allowing a credential registered for `example.com` to mint or sign keys for `www.example.com` and broaden identity binding checks across sibling hosts.

### Description

- Change allowed-domain derivation in `packages/api/src/routes/federation.ts` to record the exact redirect-uri hostname lowercased via `new URL(uri).hostname.toLowerCase()` and to authorize requests using `domain.trim().toLowerCase()` instead of `canonicalFederationHost`.
- Require exact lowercased hostname equality for signing host checks by deriving `keyIdHost` with `new URL(keyId).hostname.toLowerCase()` before comparing to the credential allow-list.
- Preserve `www.` as a distinct authority in federated-user resolution by normalizing the asserted domain to `domain.trim().toLowerCase()` in `packages/api/src/routes/users.ts` and replacing `isSameFederationHost` comparisons with exact hostname equality for username↔domain and actor↔domain binding checks.
- Update regression tests in `packages/api/src/routes/__tests__/federation.test.ts` and `packages/api/src/routes/__tests__/usersResolve.test.ts` to assert that bare/`www.` sibling hosts are rejected unless authority is established via the existing bridge trust or WebFinger paths, and remove unused `canonicalFederationHost` imports where applicable.

### Testing

- Ran `git diff --check` and inspected diffs to ensure no whitespace/errors were introduced and the intended files were modified as expected.  
- Attempted to run the targeted Jest tests with `bun run --filter @oxyhq/api test -- --runInBand src/routes/__tests__/federation.test.ts src/routes/__tests__/usersResolve.test.ts`, but tests could not be executed because the environment did not have `jest` installed (command failed).  
- Attempted `bun install --frozen-lockfile` to install dependencies required to run the test suite, but dependency resolution failed due to HTTP 403 responses from the package registry.  
- Verified working-tree status with `git status --short --branch` and confirmed the modified files are present in the repository for review.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a712450cbac8323befd5a18a741c83b)

### DIFF
--- a/packages/api/src/routes/__tests__/federation.test.ts
+++ b/packages/api/src/routes/__tests__/federation.test.ts
@@ -477,7 +477,7 @@ describe('POST /federation/sign', () => {
     expect(mockSignWithKeyId).not.toHaveBeenCalled();
   });
 
-  it('accepts a www. keyId host when the credential is registered for the bare domain', async () => {
+  it('rejects a www. keyId host when the credential is registered for the bare domain', async () => {
     const keyId = `https://www.${MENTION_DOMAIN}/ap/users/bob#main-key`;
 
     const res = await requestJson('POST', '/federation/sign', {
@@ -485,8 +485,8 @@ describe('POST /federation/sign', () => {
       signingString: SIGNING_STRING,
     });
 
-    expect(res.status).toBe(200);
-    expect(mockSignWithKeyId).toHaveBeenCalledWith(keyId, SIGNING_STRING);
+    expect(res.status).toBe(403);
+    expect(mockSignWithKeyId).not.toHaveBeenCalled();
   });
 });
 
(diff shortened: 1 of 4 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (this branch is 247 commits behind). **Rejecting the patch.**

This replaces `canonicalFederationHost()` with inline `.trim().toLowerCase()` at five call sites in `federation.ts` and `users.ts`, and `isSameFederationHost()` with `!==`. That is the one thing `packages/federation/src/apUri.ts` documents as the failure mode:

> A consumer that keeps its own copy of the rule (a moderation blocklist, a transparency page, a content purge) is keeping a second opinion about which hosts are which, and the moment the two drift the consumer acts on domains the engine never refused.

Concretely, the helper strips one leading `www.`; the inline form does not. After this patch, `loadAllowedDomains` would register `www.app.example` where `createDomainPolicy` — the gate every inbound activity and actor fetch passes — canonicalises to `app.example`. The authorisation check and the blocklist would then disagree about the same host, which is strictly worse than the current behaviour.

If the goal is genuinely exact (non-`www`-folding) comparison at the federation-credential boundary, that is a policy change and belongs in `apUri.ts` as a second explicitly-named function, applied everywhere the engine compares hosts — not as five hand-rolled copies at the call sites.

Closing. The `keyId` host binding and the `usernameDomain`/`domain` agreement checks the PR touches already exist on `main` via the shared helpers.

PR #907, opened 2026-08-10
fix(api): revoke stale background credentials on session replacement

NateIsern:
### Motivation

- Close an authorization gap where a background credential bound to an account could still mint tokens after that account's session was replaced on the same device.
- Ensure background credentials are scoped to the session state that provisioned them so delegated/managed accounts on shared devices cannot be hijacked by stale credentials.

### Description

- When re-adding/replacing an account in `addAccount`, clear the stored background credential fields so a credential issued under the previous session cannot follow the account into a new session by account id match in `packages/api/src/services/deviceSession.service.ts`.
- Add a regression test that provisions a background credential, replaces the account's session, asserts the stored background fields are cleared, and verifies `mintFromBackgroundSecret` returns `background_credential_invalid` in `packages/api/src/services/__tests__/deviceSession.service.test.ts`.
- Preserve existing behaviour for unrelated accounts and idempotent same-session re-registers; only the case that replaces an account's `sessionId` triggers clearing the background credential.

### Testing

- Ran `git diff --check` and repository status checks to validate the patch applied cleanly and the working tree is updated (success).
- Added a focused regression test in `deviceSession.service` suite, but running the repo's test script (`bun run test` / `jest`) failed in this environment because test dependencies are not installed and the `jest` binary was not available (could not execute unit test runner here).
- CI or a developer machine with node deps installed should run `bun run test -- --runInBand src/services/__tests__/deviceSession.service.test.ts` to verify the new test and the full service suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7130c5575083238342e766954fb45c)

### DIFF
--- a/packages/api/src/services/__tests__/deviceSession.service.test.ts
+++ b/packages/api/src/services/__tests__/deviceSession.service.test.ts
@@ -246,6 +246,25 @@ describe('addAccount', () => {
     expect(rows.find((r) => r.accountId === a1)?.sessionId).toBe('s-new');
   });
 
+  it('revokes a background credential when its account session is replaced', async () => {
+    const device = deviceId();
+    const a1 = await account();
+    await deviceSessionService.addAccount(device, { accountId: a1, sessionId: 's-old' });
+    mockGetAccessToken.mockResolvedValue({ accessToken: 'jwt-old', expiresAt: new Date() });
+    const credential = await deviceSessionService.issueBackgroundCredential(device, a1);
+
+    await deviceSessionService.addAccount(device, { accountId: a1, sessionId: 's-new' });
+
+    const stored = await storedDevice(device);
+    expect(stored.backgroundSecretHash).toBeNull();
+    expect(stored.backgroundSecretAccountId).toBeNull();
+    expect(stored.backgroundSecretExpiresAt).toBeNull();
+    expect(
+      await deviceSessionService.mintFromBackgroundSecret(device, credential?.secret as string)
+    ).toEqual({ ok: false, reason: 'background_credential_invalid' });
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **Premise still live, but the function it patches has been rewritten and the six-line insertion needs re-siting, not rebasing.**

The finding holds: `addAccount` still does not clear the background credential when it replaces the session for an account that owns it. `backgroundSecretAccountId` and `clearedBackgroundCredentialFields()` both still exist (`deviceSession.service.ts:230`, `:2102`), and the update in `addAccount` sets only

```ts
...(await this.resolveActiveFields(tx, current.id, nextActiveAccountId)),
revision: sql`${deviceSessions.revision} + 1`,
```

so a background credential provisioned against the previous session follows the account across a re-authentication — which is exactly what this PR argues it must not do.

Why I stopped short of resolving it myself: the surrounding code is no longer the code this diff was written against. `addAccount` now works over `deviceAccountContexts` with a principal/account split, has a "case 2" branch that moves an account to a *different operator*, calls `pruneOrphanPrincipals`, and defers `deactivateSession(displacedSessionId)` outside the transaction. The patch's condition —

```ts
existing && current.backgroundSecretAccountId === input.accountId
```

— was written when "existing" meant one thing. Under the principal split it needs deciding whether the credential should also be cleared when the account stays but its *operator* changes (case 2), which reads like the same argument the PR makes and is not covered by the condition as written.

That is a judgement about device-credential revocation semantics, and I would rather it come from you than from me guessing at the new branch. Once the condition is settled the change is small and #929 is unaffected by it.

NateIsern:

PR #908, opened 2026-08-10
fix(api): scope moderation reversals to emitter credential

NateIsern:
### Motivation
- Prevent a service credential from reversing moderation effects produced by another credential when `decisionId` values collide by ensuring reversals are scoped to the verified emitter credential.

### Description
- Forward the authenticated service credential id from the route into the reversal operation by passing `req.serviceApp!.credentialId` to `reverseModerationDecision` in `packages/api/src/routes/moderationReputation.routes.ts`.
- Restrict the reversal lookup to effects owned by the requesting credential by adding `findEffectsByDecisionForCredential(decisionId, decisionRevision, emitterCredentialId)` and using it in `reverseModerationDecision` in `packages/api/src/services/moderationReputation.service.ts`.
- Propagate credential scoping into reconciliation calls that call `reverseModerationDecision` so superseded revisions are reversed only for their emitter credential in `packages/api/src/services/moderationReputation.service.ts`.
- Update tests and test fixtures in `packages/api/src/services/__tests__/moderationReputation.service.test.ts` and `packages/api/src/routes/__tests__/moderationReputationAuthz.test.ts` to create a service credential, pass `emitterCredentialId` through world context, and add a regression asserting a credential cannot reverse another credential's colliding `decisionId`.

### Testing
- Attempted to run the package tests via `bun run test --runInBand` for the modified suites, but the test runner (`jest`) was not present in the execution environment and the test command failed; no test suites were executed successfully.
- Static checks performed: `git diff --check` produced no whitespace errors on the working tree changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a713076b4e08323b22d69b676040452)

### DIFF
--- a/packages/api/src/routes/__tests__/moderationReputationAuthz.test.ts
+++ b/packages/api/src/routes/__tests__/moderationReputationAuthz.test.ts
@@ -347,7 +347,12 @@ describe('POST /reputation/moderation/effects/reverse', () => {
       points: 9999,
     });
     expect(res.status).toBe(200);
-    expect(mockReverse).toHaveBeenCalledWith('dec_1', 1, 'Appeal accepted');
+    expect(mockReverse).toHaveBeenCalledWith(
+      'dec_1',
+      1,
+      'Appeal accepted',
+      '64bbbbbbbbbbbbbbbbbbbbbb'
+    );
   });
 
   it('requires a reason — a reversal without one is unexplainable', async () => {
--- a/packages/api/src/routes/moderationReputation.routes.ts
+++ b/packages/api/src/routes/moderationReputation.routes.ts
@@ -257,7 +257,8 @@ router.post(
     const result = await moderationReputationService.reverseModerationDecision(
       req.body.decisionId,
       req.body.decisionRevision,
-      req.body.reason
+      req.body.reason,
+      req.serviceApp!.credentialId
     );
 
     const dto: ReverseModerationEffectResult = {
--- a/packages/api/src/services/__tests__/moderationReputation.service.test.ts
+++ b/packages/api/src/services/__tests__/moderationReputation.service.test.ts
@@ -210,6 +210,16 @@ async function makeWorld(options: { globalEffects?: boolean } = {}): Promise<Wor
   const bindingId = await makeBinding(applicationId, subjectId);
   const policyVersion = await makePolicy();
   const emitterApplicationId = await makeApplication();
+  const [credential] = await getDb()
+    .insert(applicationCredentials)
+    .values({
+      applicationId: emitterApplicationId,
+      name: 'emitter credential',
+      publicKey: `oxy_dk_${uniqueId()}`,
+      type: 'service',
+      environment: 'development',
+    })
+    .returning({ id: applicationCredentials.id });
 
   mockResolveBindingProof.mockResolvedValue({ ok: true, binding: { id: bindingId } });
 
@@ -219,7 +229,7 @@ async function makeWorld(options: { globalEffects?: boolean } = {}): Promise<Wor
     bindingId,
     policyVersion,
     emitterApplicationId,
-    context: { emitterApplicationId },
+    context: { emitterApplicationId, emitterCredentialId: credential.id },
   };
 }
 
@@ -568,7 +578,8 @@ describe('DoD: an accepted appeal compensates the points and removes the active 
     const result = await moderationReputationService.reverseModerationDecision(
       event.decisionId,
       1,
-      'Appeal accepted: the material was quoted criticism, not abuse'
+      'Appeal accepted: the material was quoted criticism, not abuse',
+      world.context.emitterCredentialId!
     );
 
     expect(result.idempotent).toBe(false);
@@ -609,13 +620,15 @@ describe('DoD: an accepted appeal compensates the points and removes the active 
     await moderationReputationService.reverseModerationDecision(
       event.decisionId,
       1,
-      'Appeal accepted'
+      'Appeal accepted',
+      world.context.emitterCredentialId!
     );
 
     const again = await moderationReputationService.reverseModerationDecision(
       event.decisionId,
       1,
-      'Appeal accepted'
+      'Appeal accepted',
+      world.context.emitterCredentialId!
     );
 
     expect(again.idempotent).toBe(true);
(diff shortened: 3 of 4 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f`. **Premise live, shape right — one defect blocks the merge.**

The finding holds: `reverseModerationDecision` looks effects up by `(decisionId, decisionRevision)` only, and decision ids are chosen by the emitting app, so app A can reverse app B's moderation effect by naming its id. Scoping the lookup to `req.serviceApp.credentialId` is the right fix.

The defect is in the supersede path:

```ts
await this.reverseModerationDecision(
  effect.decisionId,
  effect.decisionRevision,
  `Superseded by revision ${latestRevision}`,
  effect.credentialId ?? ''
);
```

`moderationEffects.credentialId` is nullable — `packages/api/src/db/schema/moderationEffects.ts:123` declares `text().references(...)` with no `.notNull()`. For an effect with a NULL `credentialId` this passes `''`, `findEffectsByDecisionForCredential` matches nothing, and the call throws `NotFoundError('No moderation effect exists for that decision revision')` — so superseding silently stops reversing exactly the rows that predate credential tagging.

Either make the credential filter optional on the internal path (the supersede call already knows the effect it is reversing, so it does not need the authorization scope), or handle NULL explicitly rather than coercing it to a string that can never match.

Leaving open for that fix; happy to merge once it is in.

NateIsern:
Update: I started implementing the fix and stopped, because the column this needs to key on does not hold what its schema says it holds. That changes what this PR should do, so it is worth writing down before anyone rebases it.

## What I found

`moderationEffects.applicationId` documents itself as the emitter:

> **The emitting application, resolved from its credential — never the body.**
>
> `RESTRICT`, as the deferred-FK ledger decided before `applications` landed: the emitter is part of the provenance an effect exists to record, so deleting it must FAIL rather than erase or orphan the audit trail of what it caused.

The insert writes the reportee into it:

```ts
const [createdEffect] = await tx
  .insert(moderationEffects)
  .values({
    ...
    applicationId: reportedApplicationId,        // from the request BODY
    credentialId: context.emitterCredentialId,   // from the service token
```

and `reportedApplicationId` is `String(event.reportedApplicationId)` — read straight off the event body in `readEventIdentifiers`.

`context.emitterApplicationId` is *validated as present* (`applyModerationDecision` rejects a context without one) and then **never stored**.

The sibling table makes the mix-up legible. `conductStrikes.applicationId` documents the opposite meaning and is written with the same value:

> Which application's **report** started the incident. `SET NULL` … this is attribution and the column is already nullable, so a strike stands on its own decision and an unattributed one is a real state rather than a corrupted one.

Same column name, opposite semantics, and `moderationEffects` got the strikes table's value. One of the two is wrong; I cannot tell which from here.

## Why this blocks the PR

Three keys, none of them currently usable as-is:

- **`credentialId`** — what this PR scopes on. It is the only stored identity actually resolved from the token, but the schema makes it nullable *on purpose*: "a rotated-out credential may legitimately be removed without invalidating the effect", and "NULL already means 'not recorded' on rows predating the field". Scoping on it means an application cannot reverse **its own** effect after a routine credential rotation, and cannot reverse anything written before the column existed. That is the same defect as the `effect.credentialId ?? ''` I flagged earlier, just moved to the front door.
- **`applicationId`** — NOT NULL and RESTRICT, so it is shaped like the right key, but it currently holds the reportee. Scoping on it would let the *reported* application reverse the decision against it, which is worse than the bug being fixed.
- **`emitterApplicationId`** — the correct value, present on the context, never persisted.

I tried the `applicationId` route and the existing suite caught it immediately (`No moderation effect exists for that decision revision` on every reversal, because the fixtures' emitter is not the reportee). That is the gate doing its job.

## What I think the sequence is

1. Decide what `moderationEffects.applicationId` means. If the schema comment is right, the insert is wrong and the emitter should be stored — a migration plus a backfill, since existing rows hold the reportee.
2. Then scope the reversal on whichever column ends up holding the emitter, with `credentialId` as provenance detail only.

Doing step 2 first — which is what this PR does — bakes in a lookup key the schema explicitly says may go NULL under you.

Not touching it further. The finding in this PR is real and worth keeping: `reverseModerationDecision` is still keyed on `(decisionId, decisionRevision)` alone, and `decisionId` is the emitter's own identifier, so one application can still reverse another's consequence by naming its id.

NateIsern:
Resolved the ambiguity I flagged above. **The comment was wrong, not the code** — so there is no migration or backfill in the way, but the blocker for this PR stands and is now precisely stated.

`moderation_effects.application_id` holds the **reported** application. Every writer sends `reportedApplicationId` (ledger, strike, effect), the single reader copies it into `conduct_strikes.application_id`, and both sibling columns document that same meaning. The "emitting application… never the body" sentence came from the original Mongo model in `624ed4e50` and was never true. Corrected in #1191.

I also confirmed there is **no vulnerability** in the body-supplied value: `resolveBindingProof` requires `(bindingProofId, applicationId)` to match a live binding, bindings are created only by the application they name (`POST /bindings` takes `applicationId` from the credential plus the subject's own token), and binding ids are never echoed by a read path. Naming another application fails `no_binding_proof`.

## What that means for this PR

The emitter is **not persisted**. `applyModerationDecision` validates `context.emitterApplicationId` and then drops it; the only emitter identity on a row is `credential_id`, which the schema documents as legitimately NULL after a rotation and on pre-column rows.

So scoping the reversal on `credential_id` — what this PR does — fails in both directions:

- an application cannot reverse **its own** effect after a routine credential rotation;
- nothing written before the column existed can be reversed at all.

And scoping on `application_id` instead, which I tried, lets the **reported** application reverse the decision against it. Worse than the bug.

## The order I would suggest

1. Persist the emitter. `context.emitterApplicationId` is already validated and in hand — it needs a column (`emitter_application_id`, `NOT NULL`, `RESTRICT`) plus a backfill decision for existing rows, where the best available emitter is `credential_id` → its application, and NULL where the credential is gone.
2. Then scope the reversal on that column, with `credential_id` staying provenance detail.

Step 2 without step 1 is what this PR is, and it bakes in a lookup key the schema explicitly says may go NULL underneath it.

The finding remains valid and unfixed: `reverseModerationDecision` is still keyed on `(decisionId, decisionRevision)` alone, and `decisionId` is the emitter's own identifier.

*(Edited: an earlier version of this comment lost four identifiers to shell expansion.)*

PR #909, opened 2026-08-10
fix(api): revoke stale official redirect URIs

NateIsern:
### Motivation

- A security regression made official-app reconciliation and seeding preserve existing redirect URIs (union), which keeps stale or attacker-controlled callbacks trusted for first-party apps and expands credentialed CORS trust.
- The change restores the authoritative/converging behavior for redirect allowlists so non-canonical callbacks are revoked and the OAuth/CORS trust boundary is tightened.

### Description

- Change `computeOfficialRedirectUriRepair` in `packages/api/src/utils/redirectUris.ts` to converge the allowlist to the canonical `websiteUrl` origin (return the single canonical origin) and to treat a single-origin allowlist that already equals the canonical origin as a no-op.  
- Make the seed plan authoritative for redirect URIs in `packages/api/src/scripts/seedOxyApplicationsPlan.ts` by using the seed's `redirectUris` (replace instead of union) while keeping additive unions for `scopes` and `capabilities`.  
- Update unit tests to cover the new behavior and add a regression test that stale redirect URIs are revoked in the seed plan (`packages/api/src/utils/__tests__/redirectUris.test.ts` and `packages/api/src/scripts/__tests__/seedOxyApplicationsPlan.test.ts`).  
- Small test and helper cleanup: removed the non-authoritative `unionRedirectUris` usage for the seed/official-repair paths and adapted expectations accordingly.

### Testing

- Attempted to run the updated unit tests `src/utils/__tests__/redirectUris.test.ts` and `src/scripts/__tests__/seedOxyApplicationsPlan.test.ts`, but the package `test` script could not execute in this environment because `jest` was not available (`command not found`), so the tests were not executed here.  
- Performed repository checks (`git diff --check`) and validated the working tree and diffs locally; the changes were committed in the workspace.  
- Recommend running the package tests locally or in CI with `bun run test` (or `yarn/npm run test`) inside `packages/api` to verify all tests pass in a fully provisioned dev environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71357176e08323a9497c7806d336f2)

### DIFF
--- a/packages/api/src/scripts/__tests__/seedOxyApplicationsPlan.test.ts
+++ b/packages/api/src/scripts/__tests__/seedOxyApplicationsPlan.test.ts
@@ -78,6 +78,19 @@ describe('computeSeedApplicationPlan', () => {
     expect(plan.changes).toEqual([]);
   });
 
+  it('revokes redirect URIs which are absent from the canonical seed', () => {
+    const current = {
+      ...commonsMissingCapability(),
+      redirectUris: ['commons://', 'oxycommons://', 'https://stale.example/callback'],
+      capabilities: ['identity:approval'],
+    };
+
+    const plan = computeSeedApplicationPlan(current, COMMONS_TARGET);
+
+    expect(plan.changes.map((change) => change.field)).toEqual(['redirectUris']);
+    expect(plan.desired.redirectUris).toEqual(['commons://', 'oxycommons://']);
+  });
+
   it('applySeedApplicationPlan writes exactly the fields the plan reported', () => {
     const plan = computeSeedApplicationPlan(commonsMissingCapability(), COMMONS_TARGET);
     const record = mutableRecord(commonsMissingCapability());
--- a/packages/api/src/scripts/seedOxyApplicationsPlan.ts
+++ b/packages/api/src/scripts/seedOxyApplicationsPlan.ts
@@ -119,7 +119,10 @@ export function computeSeedApplicationPlan(
     isOfficial: true,
     isInternal: target.type === 'internal',
     ownerAccountId: target.ownerAccountId,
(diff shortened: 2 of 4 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **Closing: `main` has since decided the opposite, deliberately, and this would break path-specific callbacks.**

`packages/api/src/utils/redirectUris.ts` on `main` now opens with the policy:

> Official-app reconciliation may seed an empty allowlist, but must never broaden an explicitly configured allowlist.

and on the repair function:

> Existing entries are security-sensitive exact callback URIs and must never be changed or supplemented by this background repair.

with a test pinning it — *"preserves path-specific callbacks instead of adding their bare origin"*, asserting `null` for `['https://app.example.com/oauth/callback', 'https://staging.example.com/oauth/callback']`.

This PR converges that same allowlist to the declared website origin, and its own test asserts `['https://oxy.so', 'https://stale.example/callback']` → `['https://oxy.so']`. Applied to the case `main` pins, an official app whose callback is `https://app.example.com/oauth/callback` would end up with `https://app.example.com` — its exact callback path revoked and replaced by a **broader** bare origin. That is a worse authorization boundary than the one it removes, and it breaks the app's OAuth flow.

## The half that might still be worth doing

The other file, `seedOxyApplicationsPlan.ts`, is a different question and I do not think it is settled by the above. Making `redirectUris` authoritative *in the seed* — where the list is written down in the registry, the workflow is `workflow_dispatch` only, and `dry_run` defaults to `true` so an operator sees the plan before anything is written — is defensible in a way the background repair is not. Today it unions, so a retired callback on an official app is never revoked by reconciliation.

If you want that, it should be its own PR against `seedOxyApplicationsPlan.ts` alone, leaving `redirectUris.ts` as `main` has it. Closing this one.

PR #910, opened 2026-08-10
fix(api): tie OAuth origin verification to actual request origin

NateIsern:
### Motivation
- Close a security gap where OAuth-bound session creation used the relying-party redirect URI to skip the trusted-app origin proof and mark requests as `originVerified`, enabling attacker-created OAuth sessions for trusted apps after victim approval.
- Preserve the UX requirement that the relying-party origin remains visible in the approval UI while ensuring it cannot be used as proof that the request actually originated from that origin.

### Description
- Change origin verification in `packages/api/src/routes/auth.ts` to compute `originVerified` from the actual request Origin (`requestOriginHeader`) rather than from the OAuth `boundOrigin` derived from the redirect URI, and update the explanatory comments accordingly.
- Keep `boundOrigin` (the relying-party redirect origin) recorded for display during approval but explicitly prevent it being treated as the requester proof.
- Add regression tests in `packages/api/src/routes/__tests__/authSessionOAuthRequest.test.ts` that assert IdP-originated requests keep the displayed RP but are unverified, attacker-originated OAuth requests are not verified, and genuine requests from a registered RP origin are verified.

### Testing
- Ran `git diff --check` to validate the patch formatting and it passed. 
- Attempted to run the modified test file with `bun run test --runInBand src/routes/__tests__/authSessionOAuthRequest.test.ts`, but execution was blocked in this environment because Jest and repository dependencies are not installed (environment limitation). 
- Attempted `bun install --frozen-lockfile` to install dependencies, but it failed due to the npm registry returning HTTP 403 from this environment (environment limitation). 
- Unit tests were added to cover the regressions, but full automated execution could not be completed here due to the dependency install limitation.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a71396f2d8c8323996b765e2af443a9)

### DIFF
--- a/packages/api/src/routes/__tests__/authSessionOAuthRequest.test.ts
+++ b/packages/api/src/routes/__tests__/authSessionOAuthRequest.test.ts
@@ -376,6 +376,41 @@ describe('POST /auth/session/create — the OAuth binding', () => {
     );
 
     expect(res.status).toBe(200);
+    const row = await stored(sessionToken);
+    expect(row.boundOrigin).toBe('https://rp.example');
+    expect(row.originVerified).toBe(false);
+  });
+
+  it('does not verify a trusted OAuth app from an attacker origin', async () => {
+    const { clientId } = await client({ isOfficial: true, type: 'first_party' });
+    const sessionToken = token();
+
+    const res = await post(
+      '/auth/session/create',
+      { sessionToken, clientId, oauth: oauthBody() },
+      { origin: 'https://evil.example' },
+    );
+
+    expect(res.status).toBe(200);
+    const row = await stored(sessionToken);
+    // Keep showing the intended relying party, but do not turn its registered
+    // redirect URI into proof that this request actually came from that party.
+    expect(row.boundOrigin).toBe('https://rp.example');
+    expect(row.originVerified).toBe(false);
+  });
+
+  it('verifies a trusted OAuth app when the request origin is registered', async () => {
+    const { clientId } = await client({ isOfficial: true, type: 'first_party' });
+    const sessionToken = token();
+
+    const res = await post(
+      '/auth/session/create',
+      { sessionToken, clientId, oauth: oauthBody() },
+      { origin: 'https://rp.example' },
+    );
+
+    expect(res.status).toBe(200);
+    expect((await stored(sessionToken)).originVerified).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (247 behind; now conflicts — `main` rewrote this block around `registeredOrigins` / `matchesRegisteredOrigin`, and `19cec8119` fixed a neighbouring `Origin: null` proof bypass). **The finding survives that rewrite. The patch does not work.**

## The finding is still live

On `main`:

```ts
const boundOrigin = oauthContext
  ? originFromRedirectUri(oauthContext.redirectUri) ?? requestOriginHeader
  : requestOriginHeader;
...
const originVerified =
  isTrustedApplication(resolvedApp) &&
  !!boundOrigin &&
  matchesRegisteredOrigin(registeredOrigins, boundOrigin);
```

For an OAuth-bound request `boundOrigin` is derived from `redirect_uri`, and `redirect_uri` was exact-matched against the app's registration upstream — so it is *by construction* a registered origin and `matchesRegisteredOrigin` is trivially true. The flag therefore proves nothing about where the request came from, while its own comment still claims:

> True ONLY when a platform-trusted Application proved it is running on one of its OWN registered redirect origins.

Anyone who can start an OAuth flow with a trusted app's public `client_id` and its registered `redirect_uri` gets `originVerified: true` on the approval screen. Both values are public. That is the anti-phishing signal the approver is being asked to trust.

## Why the patch is not the fix

Switching the check to `requestOriginHeader` makes it accurate but useless for the OAuth path: the request to `/session/create` comes from the IdP shell (`auth.oxy.so`), which is not an RP redirect origin, so `originVerified` would be **false on every legitimate OAuth sign-in** and Commons would warn the approver each time. A warning that fires on the normal path is worse than no warning — it trains people to click through.

## What would work

The two meanings need separating, because `main` deliberately wants the RP origin *displayed*:

- Keep showing the RP redirect origin — that part of `boundOrigin` is intentional and documented.
- Make `originVerified` mean only what it says: a trusted app proved *the request* came from a registered origin. For OAuth-bound sessions it is then not applicable, and the approval UI should render the OAuth case as its own state ("authorising <RP>, redirect verified") rather than reusing the anti-phishing boolean.

Either way the current comment must stop claiming a proof the OAuth branch does not make.

Leaving open so the finding is not lost. Not merging this diff.

NateIsern:

PR #911, opened 2026-08-10
fix(federation): bound remote ActivityPub response bodies

NateIsern:
### Motivation
- Prevent an unauthenticated remote federated server from exhausting memory/CPU by returning extremely large actor, collection, or error responses during actor resolution.

### Description
- Add bounded read helpers `readBoundedResponseBody` and `readBoundedJson` that check `Content-Length`, stream the response via `ReadableStream` readers, enforce separate decompressed-size caps, cancel streams that exceed limits, and validate JSON shapes.
- Apply bounded reads to actor error logging and WebFinger-fallback error logging, replace unbounded `res.json()` for the actor document with `readBoundedJson(res, ACTOR_BODY_MAX_BYTES)`, and replace collection `res.json()` with `readBoundedJson(res, COLLECTION_BODY_MAX_BYTES)` so collection counts cannot trigger unbounded parsing.
- Introduce size caps constants (`ACTOR_BODY_MAX_BYTES`, `COLLECTION_BODY_MAX_BYTES`, `ERROR_BODY_MAX_BYTES`) and ensure oversized collection responses preserve the existing fail-soft behavior by returning zero counts.
- Add regression tests in `packages/federation/src/__tests__/actorResolver.test.ts` that simulate streaming responses to verify stream cancellation, early `Content-Length` rejection, oversized actor bodies, oversized collection responses, and bounded error-body handling.

### Testing
- Built the updated resolver with `bun build packages/federation/src/node/actorResolver.ts --target=node --outfile=/tmp/actorResolver.js` which completed successfully. 
- Ran `git diff --check` to validate the working tree and file changes with no whitespace/errors reported. 
- Added Jest unit tests for the new behaviour but `bun run test --runInBand` could not be executed in this environment because the package test runner (`jest`) and other dependencies are not installed. 
- Attempted `bun install --frozen-lockfile` to populate dependencies but the environment's registry access returned HTTP 403, preventing automated test execution here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7153ffcf9483238864e8f66ad16250)

### DIFF
--- a/packages/federation/src/__tests__/actorResolver.test.ts
+++ b/packages/federation/src/__tests__/actorResolver.test.ts
@@ -23,6 +23,7 @@ function makeResolver(
     cached?: TestActor | null;
     cachedKey?: Pick<TestActor, 'uri' | 'publicKeyPem'> | null;
     blockedHosts?: string[];
+    signedFetch?: (url: string) => Promise<Response>;
   } = {},
 ) {
   const blockedHosts = overrides.blockedHosts ?? ['spam.example'];
@@ -37,6 +38,7 @@ function makeResolver(
     federationEnabled: true,
     signedFetch: async (url) => {
       signedFetchCalls.push(url);
+      if (overrides.signedFetch) return overrides.signedFetch(url);
       // A 404 ends `fetchRemoteActor` without exercising the parse/upsert path;
       // these tests only care about WHETHER the fetch was attempted.
       return new Response(null, { status: 404 });
@@ -70,6 +72,22 @@ function makeResolver(
   };
 
   return { resolver: createActorResolver(config), findActorByUriCalls, signedFetchCalls };
+}
+
+function streamedResponse(chunks: string[], init?: ResponseInit): { response: Response; cancelled: () => boolean } {
+  let cancelled = false;
+  const encoder = new TextEncoder();
+  const response = new Response(new ReadableStream<Uint8Array>({
+    pull(controller) {
+      const chunk = chunks.shift();
+      if (chunk === undefined) controller.close();
+      else controller.enqueue(encoder.encode(chunk));
+    },
+    cancel() {
+      cancelled = true;
+    },
+  }), init);
+  return { response, cancelled: () => cancelled };
 }
 
 describe('getOrFetchActor — instance domain policy', () => {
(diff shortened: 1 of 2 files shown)


NateIsern:
Rebased onto current `main` and re-ran CI. **`Federation Tests` fails on this PR's own new tests** — it was green in August against the `main` of that day, so something in the 247 commits since changed the behaviour it asserts.

```
FAIL src/__tests__/actorResolver.test.ts
  ● fetchRemoteActor — bounded remote bodies › rejects and cancels an actor body that exceeds the decompressed-size cap
    Expected: true
    Received: false
  ● fetchRemoteActor — bounded remote bodies › bounds actor-advertised collection responses and fails the count soft
    Expected: true
    Received: false
  ● fetchRemoteActor — bounded remote bodies › caps an error body before attempting the WebFinger fallback
    Expected: true
    Received: false

Test Suites: 1 failed, 11 passed, 12 total
Tests:       3 failed, 212 passed, 215 total
```

All three assert `cancelled()` on the `streamedResponse` helper — i.e. that the oversized body's stream was actually cancelled rather than drained. The bound itself is presumably still applied (nothing else in the suite regressed); what no longer happens is the cancellation, which suggests `readBoundedResponseBody` is not being reached on those paths any more, or the reader's `cancel()` no longer propagates to the underlying stream after the rebase.

The change itself is worth having — `fetchRemoteActor` and the collection count still call `res.json()` unbounded on `main`, so a hostile instance can hand back an arbitrarily large actor document. It just needs the tests to pass before it can land.

One thing to fix while in here: `readBoundedResponseBody` returns `''` for a body-less response, and `readBoundedJson` then calls `JSON.parse('')`, which throws a `SyntaxError` rather than the graceful `null` the caller expects. `packages/api/src/services/linkPreview/boundedBody.ts` already solves the same problem for the API package; if this helper is going to exist in `packages/federation` too, the pair should agree on that edge, or the shared one should move to `@oxyhq/core/server` where both can import it.

I have rebased the branch onto `main` (force-pushed), so the run above is against today's tree. Leaving open.

NateIsern:
Rebased onto `main` and fixed the three failing tests. **The code was correct; the tests asserted something jest cannot observe.**

`cancelled()` was checking that the source stream's `cancel()` callback fired. Under jest it never does — `reader.cancel()` does not propagate to the underlying source there, while the identical construction in plain Node sets it. I proved it with a standalone repro, same helper, same reader loop:

```
A reader.cancel() -> cancelled = true      # node v24.20.0
```
```
[DBG] Response global = Response
Expected: true  Received: false            # same code, under jest
```

So the assertion was about the harness. That is why this was green in August and red now — the bound itself never stopped working, and I confirmed the reader does reach the cap by instrumenting it (`[DBG] cap excedido bytesRead= 1048577`).

The tests now assert what the bound exists for: an oversized body is not **drained**. The helper counts `pull`s, and each case offers 24 chunks totalling several times its cap. Positive control, per case, with the bounding reverted:

```
Expected: < 15
Received:   25
```

25 being every chunk plus the close. I also added the other direction — a body UNDER the cap is still read to the end — so a resolver that had broken into never reading a body at all cannot pass the three bounded cases and read as correct.

Two notes on the way through:

- Sizes are deliberately modest (3x the cap, not 10x). At 10x, the unbounded run does not fail — it exhausts the heap and dumps core, which fails the suite for a reason that hides what it measures. Run individually each still fails cleanly; the OOM only appears with all three unbounded in one process.
- Fixed the `readBoundedJson` edge I mentioned earlier: an empty body reached `JSON.parse('')` and surfaced a SyntaxError naming a column number, where the useful fact is that the instance sent nothing.

216/216 in `packages/federation`, 0 typecheck errors. Ready as far as I can tell.

PR #913, opened 2026-08-10
fix(api): require WebAuthn user verification

NateIsern:
### Motivation

- A recent change downgraded WebAuthn flows to allow possession-only ceremonies and then minted full sessions unconditionally, which made credentials usable without PIN/biometric verification and opened a high-severity authentication risk.

### Description

- Enforce local user verification in registration and authentication options by setting `userVerification: 'required'` in the WebAuthn option generation paths in `packages/api/src/routes/webauthn.ts`.
- Require user verification server-side by passing `requireUserVerification: true` to `verifyRegistrationResponse` and `verifyAuthenticationResponse` in `packages/api/src/routes/webauthn.ts` so assertions without UV are rejected.
- Update regression tests to reflect the new policy by asserting `userVerification: 'required'` in `packages/api/src/routes/__tests__/webauthnRegister.test.ts` and `packages/api/src/routes/__tests__/webauthnLogin.test.ts`, and remove tests that expected possession-only enrollments/logins to succeed.
- Tidy a couple of comments to reflect that `userVerified` remains a telemetry/assurance bit and is not a cryptographic attestation.

### Testing

- Ran `git diff --check` and repository text searches for `userVerification` / `requireUserVerification` to validate the edits, which succeeded.
- Updated tests were validated by static inspection (`rg`), but running the package's Jest suites with `bun run test` could not be executed in this environment because the test runner/dependencies are not installed, so full automated test execution was not performed.
- Changes were committed locally with message `fix(api): require WebAuthn user verification`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a715ccec35083239539565b1dba1c45)

### DIFF
--- a/packages/api/src/routes/__tests__/webauthnLogin.test.ts
+++ b/packages/api/src/routes/__tests__/webauthnLogin.test.ts
@@ -258,7 +258,7 @@ describe('POST /webauthn/login/options', () => {
     // The user's real credential id is surfaced so a non-discoverable hardware key
     // can be invoked by the browser.
     expect(opts.allowCredentials).toEqual([{ id: credentialID, transports: ['usb', 'nfc'] }]);
-    expect(opts.userVerification).toBe('preferred');
+    expect(opts.userVerification).toBe('required');
 
     // The stored challenge is bound to the resolved account and is SPENDABLE.
     const stored = await storedChallenge(currentChallenge);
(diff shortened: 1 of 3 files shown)


NateIsern:
Closing: **this reverts a deliberate decision, and the incident that produced it is written into the code being changed.**

The PR flips `userVerification` to `'required'` and `requireUserVerification` to `true` in all four places. The comment immediately above the first one records what happened last time:

> `required` here is exactly what made a Google Titan fail Chrome's "device can't be used with this site" gate.

That is the owner possession-credential policy, and the comments state the reasoning in full: a UV-capable authenticator (Face ID, Windows Hello, FIDO2-with-PIN) **still performs user verification unchanged** under `'preferred'`; only a UV-incapable U2F/CTAP1 key falls back to presence-only. So `required` buys nothing for the authenticators that already verify, and locks out the ones that cannot — which is the failure that was already hit and fixed.

The assurance level is not lost either: each ceremony's UV bit is persisted on the credential (`userVerified`), refreshed on every assertion, precisely so a future step-up can gate on UV-backed credentials without excluding hardware keys from sign-in.

Worth keeping in view: `main`'s own caveat says `userVerified` is authenticator-self-asserted and not attestation-proven, so if a hard UV boundary is genuinely wanted it needs attestation, not this flag and not this flip. That would be a fresh design, not a revert.

PR #915, opened 2026-08-10
fix(create-oxy-app): isolate generated AWS deploy roles

NateIsern:
### Motivation

- The generated GitHub Actions deploy workflow used a shared OIDC role and a user-controlled `APP` slug, allowing a generated repo to overwrite another app's SSM parameters, push to its ECR repo, or trigger its ECS service when the shared role was assumable.
- Reduce blast radius by enforcing per-app authorization boundaries in the scaffolded workflow and documenting least-privilege provisioning requirements.

### Description

- Updated the deploy workflow template to assume an app-scoped role name `arn:aws:iam::237343248947:role/oxy-{{APP_SLUG}}-github-deploy` instead of the shared `oxy-github-deploy` role and added a comment describing the trust/permission requirements in the template (`packages/create-oxy-app/templates/deploy/DOT_github/workflows/deploy-aws.yml`).
- Removed writes to the shared SSM namespace and long-lived AWS credential allowlist from the template by always writing secrets to the app-specific path `/oxy/$APP/$k` and removing `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/shared SSM sync behavior in the workflow.
- Added a regression test in `packages/create-oxy-app/src/__tests__/render.test.ts` that renders the workflow template and asserts the generated content uses the app-specific role and no longer references shared secrets or `/oxy/_shared/`.
- Documented the requirement to provision an app-specific OIDC role with limited permissions in the scaffolder `README` (`packages/create-oxy-app/README.md`).

### Testing

- Ran `bun run test` in `packages/create-oxy-app` and all tests passed (`18` tests total).
- Ran `git diff --check` and inspected the diff to confirm only intended changes were introduced, with no whitespace/errors reported.
- Attempted `bun run typecheck` but it is blocked in this environment due to missing Node type definitions and a TypeScript `moduleResolution=node10` deprecation error, so full typecheck was not completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a715ea4dbd8832382ca2aea15b09086)

### DIFF
--- a/packages/create-oxy-app/README.md
+++ b/packages/create-oxy-app/README.md
@@ -50,6 +50,14 @@ BloomThemeProvider → OxyProvider → ImageResolver → LocaleProvider`) with t
 `Stack` as the sole `(auth)`↔`(app)` authority, keyed on the device-first
 session.
 
+### AWS deployment setup
+
+Before enabling the generated workflow, provision its app-specific IAM role
+(`oxy-<slug>-github-deploy`). Its OIDC trust policy must match the generated
+GitHub repository exactly, and its permissions must be limited to that app's
+`/oxy/<slug>/*` SSM parameters, `oxy/<slug>` ECR repository, and `<slug>` ECS
+service. Do not share a deploy role between generated repositories.
+
 ## Oxy client registration
 
 By default the CLI offers to register an `Application` + public credential with
--- a/packages/create-oxy-app/src/__tests__/render.test.ts
+++ b/packages/create-oxy-app/src/__tests__/render.test.ts
@@ -121,3 +121,25 @@ describe('base template bunfig.toml', () => {
     expect(renderString(raw, ctx, 'bunfig.toml.tpl')).toContain('linker = "hoisted"');
   });
 });
+
+describe('AWS deploy template', () => {
+  test('uses an app-specific role and cannot overwrite shared secrets', async () => {
+    const templateFile = path.join(
+      __dirname,
+      '..',
+      '..',
+      'templates',
+      'deploy',
+      'DOT_github',
+      'workflows',
+      'deploy-aws.yml',
+    );
+    const rendered = renderString(await fs.readFile(templateFile, 'utf8'), ctx, templateFile);
+
+    expect(rendered).toContain('role/oxy-my-app-github-deploy');
+    expect(rendered).not.toContain('role/oxy-github-deploy');
+    expect(rendered).not.toContain('/oxy/_shared/');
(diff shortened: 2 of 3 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **Premise live, but the diff is stale in a way that would break every generated app.**

Still true on `main`'s `templates/deploy/DOT_github/workflows/deploy-aws.yml`:

```yaml
role-to-assume: arn:aws:iam::237343248947:role/oxy-github-deploy
...
SHARED="AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY REDIS_URL"
case " $SHARED " in *" $k "*) path="/oxy/_shared/$k";; *) path="/oxy/$APP/$k";; esac
```

Every generated repository assumes the same deploy role and writes long-lived `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` into a `/oxy/_shared/` path all of them read. One generated app's repo secrets can overwrite the credentials every other generated app runs with. The per-app role and per-app SSM path in this PR are the right answer, and `{{APP_SLUG}}` is a supported template token (`render.test.ts` covers YAML rendering that substitutes it while leaving `${{ github.sha }}` alone), so that part works.

What blocks it: `main` has since moved the allowlist from Mongo to Postgres —

```yaml
SSM_SECRET_ALLOWLIST: >-
  DATABASE_URL REDIS_URL
  OXY_SERVICE_API_KEY OXY_SERVICE_API_SECRET
  AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
```

This diff sets it to `MONGODB_URI OXY_SERVICE_API_KEY OXY_SERVICE_API_SECRET`, which would (a) reinstate `MONGODB_URI` and drop `DATABASE_URL`, leaving generated backends with no database URL synced, and (b) drop `REDIS_URL` entirely rather than just moving it off the shared path.

The rework is small: keep `DATABASE_URL REDIS_URL OXY_SERVICE_API_KEY OXY_SERVICE_API_SECRET`, drop only the two AWS key entries (a per-app OIDC role makes them unnecessary — that is the point of the change), delete the `SHARED` branch so everything lands under `/oxy/$APP/`, and keep the `{{APP_SLUG}}` role plus the README section.

Note that moving `REDIS_URL` off `/oxy/_shared/` is a real change for existing generated apps, not just for new ones — worth saying so in the README section this PR already adds.

Leaving open.

NateIsern:

PR #921, opened 2026-08-10
fix(api): constrain Homiio reputation awards

NateIsern:
### Motivation

- Close an authorization-boundary gap introduced by granting Homiio the broad `reputation:write` application scope which allowed arbitrary ledger mutations. 
- Limit Homiio's authority to only the intended lease lifecycle events and enforce idempotency to prevent misuse by service credentials that inherit app-level scopes.

### Description

- Introduces a new privileged scope `reputation:lease:write` and documents it in `packages/api/src/utils/applicationScopes.ts`. 
- Replaces Homiio's seeded `reputation:write` grant with the narrower `reputation:lease:write` in `packages/api/scripts/seed-oxy-applications.ts` so new credentials do not receive global ledger-write authority. 
- Adds route-level authorization in `packages/api/src/routes/reputation.routes.ts` that: allows `reputation:write` callers to retain existing behavior, permits `reputation:lease:write` callers only for the four lease action types (`lease_signed`, `lease_completed`, `clean_moveout`, `lease_default`) and requires `sourceActionId` for idempotency. 
- Adds a migration `packages/api/drizzle/0021_narrow_homiio_reputation_scope.sql` to update existing Homiio app rows and to update DB scope CHECK constraints to include the new scope. 
- Adds regression tests in `packages/api/src/routes/__tests__/reputationAwardAuthz.test.ts` verifying that `reputation:lease:write` rejects unrelated actions, enforces the idempotency key, and permits valid idempotent lease awards.

### Testing

- Ran `git diff --check` and validated the repo diffs and new test additions; that check succeeded. 
- Verified `packages/api/drizzle/meta/_journal.json` is valid JSON after adding the migration entry. 
- Attempted to run the updated package test file with `bun run --filter @oxyhq/api test --runTestsByPath src/routes/__tests__/reputationAwardAuthz.test.ts --runInBand` but test execution could not run in this environment because `jest` / workspace dependencies are not installed, so the new tests were added but not executed here. 
- All changes are limited to authorization, seed data, migration, and tests to minimize behaviour changes for existing privileged callers holding the original `reputation:write` scope.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7175ae70c08323904cc46ffe104b75)

### DIFF
--- a/packages/api/drizzle/0021_narrow_homiio_reputation_scope.sql
+++ b/packages/api/drizzle/0021_narrow_homiio_reputation_scope.sql
@@ -1,0 +1,15 @@
+-- Homiio reports lease lifecycle events; it must not inherit the global ledger
+-- mutation authority used by trusted platform services.
+ALTER TABLE "account_credentials" DROP CONSTRAINT "account_credentials_scopes_check";--> statement-breakpoint
+ALTER TABLE "application_credentials" DROP CONSTRAINT "application_credentials_scopes_check";--> statement-breakpoint
+ALTER TABLE "applications" DROP CONSTRAINT "applications_scopes_check";--> statement-breakpoint
+
+UPDATE "applications"
+SET "scopes" = array_append(array_remove("scopes", 'reputation:write'), 'reputation:lease:write')
+WHERE lower(btrim("name")) = 'homiio'
+  AND "scopes" @> array['reputation:write']::text[]
+  AND NOT "scopes" @> array['reputation:lease:write']::text[];--> statement-breakpoint
+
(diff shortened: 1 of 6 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **Not a stale-premise problem — a renumbering problem, plus a scope question.**

Two things to fix before this can land:

**1. The migration number is long gone.** This adds `packages/api/drizzle/0021_narrow_homiio_reputation_scope.sql`; `main` is at `0074_minor_silk_fever.sql`. It has to be renumbered to the next free slot with `_journal.json` regenerated, and the journal conflict here is that collision, not a content disagreement.

**2. The narrowing needs to match what Homiio actually awards today.** The change introduces `reputation:lease:write` and constrains it to `LEASE_SIGNED_ACTION`, `LEASE_COMPLETED_ACTION`, `CLEAN_MOVEOUT_ACTION`, `LEASE_DEFAULT_ACTION`, additionally requiring `sourceActionId`. The direction is right — a Homiio credential holding blanket `reputation:write` can award any action type to any principal, and a lease-scoped credential should not be able to. But whether those four are the complete set Homiio emits is a fact about Homiio, not about this repo, and if it emits a fifth the awards start 403ing after the swap.

Worth confirming against Homiio's emit sites before the credential is migrated off `reputation:write`, and worth keeping both scopes accepted (as this PR does) so the swap is not a flag-day.

Leaving open.

NateIsern:

PR #923, opened 2026-08-10
Fix(auth): prevent public auth from binding to unproved device sessions

NateIsern:
### Motivation
- A public authentication path accepted a client-supplied `deviceId` and stamped it onto newly-created sessions, allowing an attacker who knows another device's `deviceId` to have their sign-in mint the victim's device secret and subsequently obtain bearer tokens for the victim's active account.
- The intended security model requires proof of possession of the existing `deviceSecret` (or a bearer-authenticated caller) before reusing or rotating an existing device's secret; public bodies must not be able to select an authoritative `deviceId`.

### Description
- Stop trusting caller-provided `deviceId` in public auth paths by making `sessionCreateOptionsFromBody` ignore `deviceId` and only return non-authoritative device metadata (`deviceName`, `deviceFingerprint`).
- Remove `deviceId` from public request schemas and contracts so validation strips unproved device identifiers, including `packages/api/src/schemas/auth.schemas.ts` and `packages/contracts/src/webauthn.ts`.
- Update handlers to stop persisting or passing an untrusted `deviceId` during session creation and WebAuthn/challenge verify flows, including `packages/api/src/controllers/session.controller.ts`, `packages/api/src/routes/auth.ts`, and `packages/api/src/routes/webauthn.ts`.
- Add regression tests that assert public auth schema parsing and the session-options boundary no longer accept `deviceId` (`packages/api/src/schemas/__tests__/auth.schemas.test.ts`, `packages/contracts/src/__tests__/webauthn.test.ts`) and a unit test for `sessionCreateOptionsFromBody` in the session controller test file.
- Commit recorded as `fix(auth): reject unproved device session binding` with the implemented changes across the files above.

### Testing
- Ran `git diff --check` to validate no obvious whitespace/conflict issues; it passed.
- Added unit/contract tests covering the new validation behavior (files under `packages/api/src/schemas/__tests__` and `packages/contracts/src/__tests__`), but running the full test suite was blocked in this environment because the workspace dependencies / `jest` are not installed (attempted `bun run test` / `jest` produced `command not found`).
- Verified code edits with local repository checks (`git status`, diff/commit) and grepped code paths to ensure all public auth entry points no longer accept a caller `deviceId` at the handler/schema boundary.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a717c0111048323b564eb44093fe60e)

### DIFF
--- a/packages/api/src/controllers/__tests__/session.controller.test.ts
+++ b/packages/api/src/controllers/__tests__/session.controller.test.ts
@@ -77,7 +77,22 @@ import { userAuthMethods } from '../../db/schema/userAuthMethods';
 import { userLinkMetadata } from '../../db/schema/userLinkMetadata';
 import { users } from '../../db/schema/users';
 import type { AuthRequest } from '../../middleware/auth';
-import { SessionController } from '../session.controller';
+import { SessionController, sessionCreateOptionsFromBody } from '../session.controller';
+
+describe('sessionCreateOptionsFromBody', () => {
+  it('does not trust a public authentication body to select an existing device', () => {
+    const body = {
+      deviceName: 'Browser',
+      deviceFingerprint: 'metadata-only',
+      deviceId: 'victim-device',
+    };
+
+    expect(sessionCreateOptionsFromBody(body)).toEqual({
+      deviceName: 'Browser',
+      deviceFingerprint: 'metadata-only',
+    });
+  });
+});
 
 /** A captured `res` — the status and body the handler actually produced. */
 interface CapturedResponse {
--- a/packages/api/src/controllers/session.controller.ts
+++ b/packages/api/src/controllers/session.controller.ts
@@ -29,16 +29,11 @@ import type { SessionCreateOptions } from '../types/session.types';
 export function sessionCreateOptionsFromBody(body: {
   deviceName?: string;
   deviceFingerprint?: string;
-  deviceId?: string;
 }): SessionCreateOptions {
-  const opts: SessionCreateOptions = {
+  return {
     deviceName: body.deviceName,
     deviceFingerprint: body.deviceFingerprint,
   };
-  if (typeof body.deviceId === 'string' && body.deviceId.trim()) {
-    opts.deviceId = body.deviceId.trim();
-  }
-  return opts;
 }
 
 /**
@@ -337,7 +332,7 @@ export class SessionController {
    */
   static async verifyChallenge(req: Request, res: Response) {
     try {
-      const { publicKey, challenge, signature, timestamp, deviceName, deviceFingerprint, deviceId } = req.body;
+      const { publicKey, challenge, signature, timestamp, deviceName, deviceFingerprint } = req.body;
       const db = getDb();
 
       if (!publicKey || !challenge || !signature || !timestamp) {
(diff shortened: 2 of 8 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **The finding is live and serious. The patch is right for one of the two endpoints it touches and wrong for the other.**

## The finding, traced on today's `main`

`sessionService.createSession` still takes an explicit device id and uses it verbatim, ahead of everything else:

> `Precedence: deviceId > stableDeviceKey > UA/IP > random.`

and the comment names the intended caller — *"the account-switch route threading the operator's own central deviceId"*, i.e. a server-side, already-authorized caller.

But `sessionCreateOptionsFromBody` still reads it from the **request body**, and `SessionController.verifyChallenge` still destructures `deviceId` off `req.body` and passes it through (`session.controller.ts:342`, `:421`). `verifyChallenge` is public-key sign-in — it authenticates the caller as *themselves*, and nothing checks that the named device is theirs. `auth.schemas.ts` still accepts the field (`deviceIdField` on both `verifyChallengeSchema` and `authSessionCreateSchema`).

So a caller signing in with their own key can stamp any device id they know onto their session, and `finalizeDeviceLogin` then mints a device secret against it.

This also undercuts the lane split in the OAuth token exchange that I cited when closing #925 and #930: that split trusts `sharedDeviceId`, which comes from `authSessions.deviceId`, which `POST /auth/session/create` writes straight from an unauthenticated request body.

## Why I did not merge it

The two endpoints are not the same case:

- **`verifyChallenge`** — nothing in the schema or the code justifies a client-supplied device id here. Removing it is safe, and that half of this patch should land.
- **`POST /auth/session/create`** — the field carries a documented purpose: *"Originating RP device id — converges QR sign-in onto the same DeviceSession."* Deleting it means a QR sign-in mints a fresh device session instead of joining the browser's, which is what makes several accounts share one device. That is a feature regression, not a hardening.

## What the second half should be instead

The browser starting a QR flow already holds that device's `deviceSecret`. Convergence should require **proving possession** of it rather than merely naming the id — accept `deviceId` only alongside a valid secret for that device, and fall back to a fresh device session when the proof is absent or wrong. That keeps QR convergence and closes the binding.

Happy to land the `verifyChallenge` half on its own if you want the bleeding stopped first; the `/session/create` half needs the proof designed.

NateIsern:
Landed the `verifyChallenge` + WebAuthn half as #1190.

That half is safe on its own: the SDK never sent `deviceId` on those lanes (`OxyServices.verifyChallenge` has no such parameter, and no production caller builds a WebAuthn envelope with one), so removing it takes nothing away. Verified with a positive control — reverting only the controller change reproduces the pin:

```
● ignores a deviceId in the body
  Expected path: not "deviceId"   Received value: "someone-elses-device"
```

The `/auth/session/create` half stays out, for the reason in my earlier comment: its `deviceId` is documented as converging QR sign-in onto the browser's DeviceSession, so deleting it is a feature regression. That lane wants proof of possession of the device secret, not removal of the field.

Leaving this open as the tracker for that remaining half — its scope is now just `authSessionCreateSchema` and the QR convergence path.

NateIsern:

PR #925, opened 2026-08-10
fix(api): restrict OAuth device restore credentials to trusted apps

NateIsern:
### Motivation

- Prevent third-party OAuth clients from receiving a device-level restore credential (`deviceSecret`) that would allow cross-app device token minting.
- Preserve the existing first-party password handoff flow for trusted platform apps while closing the newly-introduced privilege escalation vector.

### Description

- Gate `finalizeDeviceLogin` in the OAuth token exchange behind the existing `isTrustedApplication(app)` predicate so only trusted/first-party apps receive device-first credentials.【F:packages/api/src/routes/auth.ts†L3097-L3112】
- Omit `deviceId` and `deviceSecret` from token responses for third-party clients and only include them when the app is trusted.【F:packages/api/src/routes/auth.ts†L3122-L3139】
- Add regression tests that assert third-party clients do not trigger device finalization and trusted first-party clients still receive restore credentials.【F:packages/api/src/routes/__tests__/oauthToken.test.ts†L302-L340】
- Keep the existing fail-closed behavior by exercising the device-secret mint failure path for the trusted-app route.【F:packages/api/src/routes/__tests__/oauthToken.test.ts†L716-L729】

### Testing

- ✅ Ran `git diff --check` to validate patch whitespace and basic diffs and confirmed no conflicts.
- ⚠️ Attempted to run the modified OAuth unit test via `bun run test -- --runInBand src/routes/__tests__/oauthToken.test.ts`, but the environment lacks Jest (`jest: command not found`), so the test run could not be executed here.【F:packages/api/src/routes/__tests__/oauthToken.test.ts†L300-L340】
- ✅ Added and committed the test coverage that verifies behavior for both third-party and trusted first-party clients, but full test execution must be run in CI or a developer environment with the project test dependencies installed (`npm install`/`bun install` + `npm run test` / `bun run test`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a717d1b63f0832393531c43ab0698d0)

### DIFF
--- a/packages/api/src/routes/__tests__/oauthToken.test.ts
+++ b/packages/api/src/routes/__tests__/oauthToken.test.ts
@@ -300,7 +300,7 @@ beforeEach(() => {
 });
 
 describe('POST /auth/oauth/token — RFC 6749 §5.1 success response', () => {
-  it('returns a FLAT token document (no `data` wrapper) for a PKCE public client', async () => {
+  it('returns a FLAT token document without device credentials for a third-party client', async () => {
     const res = await requestForm(pkceParams());
 
     expect(res.status).toBe(200);
@@ -310,9 +310,10 @@ describe('POST /auth/oauth/token — RFC 6749 §5.1 success response', () => {
       token_type: 'Bearer',
       expires_in: 900,
       session_id: 'sess-1',
-      deviceId: 'device-1',
-      deviceSecret: 'device-secret-1',
     });
+    expect(res.body).not.toHaveProperty('deviceId');
+    expect(res.body).not.toHaveProperty('deviceSecret');
+    expect(mockFinalizeDeviceLogin).not.toHaveBeenCalled();
     expect(res.body).not.toHaveProperty('data');
     // The zero-cookie transport hands out no refresh token.
     expect(res.body).not.toHaveProperty('refresh_token');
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing: **already fixed on `main`, by isolation rather than omission.**

`main` now splits the lane at session creation in the token exchange:

```ts
const trusted = isTrustedApplication(app);
...
  ...(trusted
    ? sharedDeviceId ? { deviceId: sharedDeviceId } : {}
    : { stableDeviceKey: `oauth:${credential.publicKey}`, application: {...} })
```

A third party no longer joins the shared device session at all — it gets an isolated per-`(user, client)` device. The code says so directly:

> What the omission was protecting against is already closed by the lane split: before this change a third party joined the SHARED device and got ITS secret, which is the credential #937 calls global. It no longer can.

> The pair names the session's OWN device — the browser's shared one for a trusted app, an isolated per-(user, client) one for a third party — so it is never a credential for a device the caller does not own.

So withholding `deviceId`/`deviceSecret` from untrusted apps is no longer the protection it was when this was written: the credential they receive is now scoped to a device only they hold. Suppressing it would just remove restore support from third-party clients without adding safety.

PR #929, opened 2026-08-10
Revoke shared device secret on account signout

NateIsern:
### Motivation

- Prevent a retained device-wide `deviceSecret` from being used to mint tokens for another account after one account signs out on a multi-account device.

### Description

- Always clear the device's rotating secret material (`secretHash`, `prevSecretHash`, `prevSecretExpiresAt`) when any account is removed from a device so the shared credential cannot be reused after signout (change in `packages/api/src/services/deviceSession.service.ts`).
- Update tests to assert single-account signout now revokes both the current and grace-window secrets and that `getStateBySecret` rejects previously issued secrets (changes in `packages/api/src/services/__tests__/deviceSession.service.test.ts`).
- Preserve existing background-credential behavior while ensuring its removal also clears the shared device secret as appropriate.

### Testing

- Ran static check `git diff --check`, which reported no whitespace/patch issues.
- Added/updated regression tests in `packages/api/src/services/__tests__/deviceSession.service.test.ts` covering signout behavior and `getStateBySecret` rejection of retained secrets; they cannot be executed end-to-end here because the environment is missing test dependencies (`jest`/workspace install), so test execution was not completed (`jest: command not found`).
- Verified repository diffs and local TypeScript/JS edits for syntactic consistency via `git status`/`git diff --stat` in the working tree.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a717fda290c832384da4c1b457c283b)

### DIFF
--- a/packages/api/src/services/__tests__/deviceSession.service.test.ts
+++ b/packages/api/src/services/__tests__/deviceSession.service.test.ts
@@ -549,19 +549,23 @@ describe('signout — device-secret cleanup', () => {
     expect(after.backgroundSecretExpiresAt).toBeNull();
   });
 
-  it('single-account signout does NOT clear the device secret (other accounts still mint with it)', async () => {
+  it('single-account signout revokes the shared device secret', async () => {
     const device = deviceId();
     const a1 = await account();
     const a2 = await account();
     await deviceSessionService.addAccount(device, { accountId: a1, sessionId: 's1' });
     await deviceSessionService.addAccount(device, { accountId: a2, sessionId: 's2' });
-    await deviceSessionService.issueDeviceSecret(device);
-    const before = await storedDevice(device);
+    const previousRetainedSecret = await deviceSessionService.issueDeviceSecret(device);
+    const retainedSecret = await deviceSessionService.issueDeviceSecret(device);
 
     await deviceSessionService.signout(device, { accountId: a1 });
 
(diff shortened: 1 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (247 behind, and it now conflicts). **Not merging: it reverses a documented deliberate behaviour, and the trade-off is yours.**

The patch makes *every* signout clear the device's shared `deviceSecret`, where `main` clears it only on signout-ALL. The comment it deletes states the reason for the current split:

> Single-account signout leaves the secret alone — other accounts on the SAME device still legitimately mint with it.

Both readings are coherent:

- **Yours:** `deviceSecret` is not account-scoped, so a signed-out account's retained copy of it can mint for whichever account remains active. That is real.
- **Current:** on a genuinely shared device, clearing it on single-account signout signs *everyone* out — one person leaving a family tablet drops the other accounts' ability to refresh.

So this is not a bug fix, it is a choice about what a shared device means. If the credential really must be per-account, the fix is to scope it (a per-account secret, or binding the mint to the account that proved it) rather than to widen the revocation and take the collateral logouts.

Related: [#907](https://github.com/OxyHQ/oxy/pull/907) covers the narrower background-credential case on session replacement and does not have this blast radius.

Leaving open pending your call.

NateIsern:

PR #930, opened 2026-08-10
fix(api): isolate OAuth sessions from shared devices

NateIsern:
### Motivation

- Close an authorization-boundary vulnerability where an OAuth token grant could be attributed to a user's shared central `DeviceSession` and thereby allow an OAuth client to operate or disrupt other accounts registered on that device.
- Ensure OAuth-issued sessions are isolated from the authorizing app's multi-account device state while preserving delegated-operator binding semantics.

### Description

- Stop threading the authorizing `deviceId` from the redeemed auth code into `sessionService.createSession` by removing the use of `exchange.code.deviceId` in the token exchange code path (`packages/api/src/routes/auth.ts`).
- Preserve delegation metadata by continuing to pass `operatedByUserId` into the newly-created OAuth session so managed-account semantics remain intact.
- Add a regression assertion in the token-exchange unit test (`packages/api/src/routes/__tests__/oauthToken.test.ts`) that the created OAuth session object does not receive a `deviceId` while still containing `operatedByUserId`.
- Commit message: `fix(api): isolate OAuth device sessions` (changes to `auth.ts` and the related test file).

### Testing

- Ran `git diff --check` which reported no whitespace or simple-diff issues.
- Updated and committed the test `packages/api/src/routes/__tests__/oauthToken.test.ts` to assert the absence of `deviceId`, but executing the test runner failed because `jest` is not installed in the execution environment (`jest: command not found`), so the modified test was not executed here.
- Manual inspection and local static checks confirm the removed device attribution and the new test assertion are present in the diff; runtime verification should be performed in CI or a developer environment with project test dependencies installed (`bun install` / `npm install` and `npm test` / `bun run test`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a718450bcb48323acc147f871f1e134)

### DIFF
--- a/packages/api/src/routes/__tests__/oauthToken.test.ts
+++ b/packages/api/src/routes/__tests__/oauthToken.test.ts
@@ -383,7 +383,7 @@ describe('POST /auth/oauth/token — RFC 6749 §5.1 success response', () => {
     expect(app.lastUsedAt).toBeInstanceOf(Date);
   });
 
-  it('threads the originating device and the delegated operator onto the session', async () => {
+  it('isolates the OAuth session from the originating device while preserving delegation', async () => {
     const org = await subject({ kind: 'organization' });
     const operator = await subject();
     mockExchangeAuthCode.mockResolvedValueOnce(
@@ -395,8 +395,9 @@ describe('POST /auth/oauth/token — RFC 6749 §5.1 success response', () => {
     expect(mockCreateSession).toHaveBeenCalledWith(
       org,
       expect.anything(),
-      expect.objectContaining({ deviceId: 'dev-shared', operatedByUserId: operator }),
+      expect.objectContaining({ operatedByUserId: operator }),
     );
+    expect(mockCreateSession.mock.calls[0]?.[2]).not.toHaveProperty('deviceId');
   });
 
   it('never serializes a protected user column into the token response', async () => {
--- a/packages/api/src/routes/auth.ts
+++ b/packages/api/src/routes/auth.ts
@@ -3070,11 +3070,6 @@ router.post(
       throw OAuthError.invalidGrant(INVALID_GRANT_DESCRIPTION);
     }
 
-    const sharedDeviceId =
-      typeof exchange.code.deviceId === 'string' && exchange.code.deviceId.trim()
-        ? exchange.code.deviceId.trim()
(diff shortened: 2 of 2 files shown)


NateIsern:
Triaged against `main` @ `c6f01a9f` (247 behind, now conflicts in `auth.ts` and `oauthToken.test.ts`). **The finding is real; the blast radius needs your call.**

The finding: the OAuth token exchange reuses `exchange.code.deviceId` as the minted session's device, so the OAuth client lands on the *authorizing app's central device session*. Through `/session/device/*` that client can then operate every account registered on that device, not just the one it was authorized for. An OAuth client is an authorization boundary and should not inherit that.

The reason I am not merging it: putting OAuth bearers on an isolated device session changes what those clients can see. Anything that today relies on the shared device session to enumerate or switch accounts after an OAuth exchange — account switching in Commons, multi-account UI backed by `/session/device/*` — would stop working, and I cannot establish from this repo alone which callers depend on it.

What would settle it:

1. Which flows call `/session/device/*` with a token minted through `POST /oauth/token`? If the answer is "none", this is safe and I will land it.
2. If some do, they need an explicit path (a device-scoped grant the approver actually sees and consents to) before the isolation lands, otherwise this is a silent capability removal for first-party surfaces.

Leaving open pending that.

NateIsern:
Correction to my earlier comment on this PR — I had not read far enough into the token-exchange handler on `main`.

**This is already fixed**, and better than the patch does it. `main` gates the shared-device reuse on trust:

```ts
const trusted = isTrustedApplication(app);
...
  ...(trusted
    ? sharedDeviceId ? { deviceId: sharedDeviceId } : {}
    : { stableDeviceKey: `oauth:${credential.publicKey}`, application: {...} })
```

A third-party OAuth client gets an isolated per-`(user, client)` device session, so it can no longer operate the other accounts on the authorizing device through `/session/device/*` — which is exactly the finding. A trusted first-party app still joins the shared device, which is the behaviour this PR would have removed and which the surrounding comment documents as intentional ("A TRUSTED application is an official Oxy app joining the shared device session, so it keeps all three unchanged").

So the questions I raised about blast radius are moot: `main` keeps the first-party capability and removes the third-party one, rather than removing both. Closing.

PR #932, opened 2026-08-10
fix(services): revoke remote sessions on sign out all

NateIsern:
### Motivation

- The public `signOutAll`/`logoutAll` contract must revoke sessions across all devices, but recent changes routed it to the device-scoped signout only, leaving remote/stolen sessions active.
- Restore the prior global revocation behavior while preserving the device-local cleanup and device-session projection semantics.

### Description

- Call the user-wide revocation endpoint via `oxyServices.logoutAllSessions(activeSessionId)` before performing the device-scoped removal `sessionClient.signOut({ all: true })` in `useAuthOperations.logoutAll`.
- Update the hook documentation to clarify that `logoutAll` combines a user-wide revoke with device-set removal, while `logout` remains device-scoped.
- Add/adjust unit-test mocks to include `logoutAllSessions` and enhance `packages/services/__tests__/context/useAuthOperations.test.tsx` with assertions that global revocation is invoked and that device teardown is skipped if the global revoke fails.
- Include `oxyServices` in the `useCallback` dependency list to satisfy React hook dependencies.

### Testing

- Ran `git diff --check` and committed the changes (local pre-commit checks passed).
- Added/updated unit tests in `packages/services/__tests__/context/useAuthOperations.test.tsx` that verify successful flow (global revoke then device sign-out) and failure flow (global revoke error prevents device sign-out and teardown).
- Attempted to install dependencies and run the package test harness (`bun install` / package tests), but dependency resolution failed due to external registry 403s so Jest tests could not be executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7190973ee4832386f51b52f640997d)

### DIFF
--- a/packages/services/__tests__/context/useAuthOperations.test.tsx
+++ b/packages/services/__tests__/context/useAuthOperations.test.tsx
@@ -11,10 +11,10 @@
  * so the test exercises the orchestration logic only — actual network, crypto,
  * and persistence belong to `@oxyhq/core` and are covered by its own tests.
  *
- * `logout` / `logoutAll` route SERVER-side revocation through a mocked
- * `SessionClient` (device-first) rather than the bearer/cookie logout
- * endpoints. A genuine FULL sign-out additionally clears the persisted refresh
- * family via `store.clear()` (`clearPersistedAuthSafe`).
+ * `logout` routes server-side revocation through a mocked `SessionClient`.
+ * `logoutAll` combines user-wide revocation with device-set removal. A genuine
+ * full sign-out additionally clears the persisted device credential via
+ * `store.clear()` (`clearPersistedAuthSafe`).
  */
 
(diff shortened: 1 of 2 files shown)


NateIsern:
Closing as a duplicate of #931. Both change the same line of `useAuthOperations.ts` in the same way — `await oxyServices.logoutAllSessions(activeSessionId)` before `sessionClient.signOut({ all: true })` — and differ only in the comment wording.

Keeping #931, whose comment states the ordering constraint explicitly (the global endpoint preserves the current session so the device-scoped cleanup can still authenticate), which is the part a reader needs.

The finding is live on `main`: `logoutAll` calls only `sessionClient.signOut({ all: true })`, which is device-scoped, so sessions on other devices survive a "sign out everywhere".

PR #935, opened 2026-08-10
Preserve application membership boundaries during account migration (Phase 4 audit-only)

NateIsern:
### Motivation

- A migration folded legacy per-application memberships into account-wide AccountMember rows, which broadened legacy app-scoped grants to every application on the owning account and created an authorization boundary regression. 
- The goal is to prevent accidental escalation of app-only roles (notably credential-management rights) while surfacing unresolved legacy memberships for an operator to remediate explicitly.

### Description

- Change Phase 4 migration to audit-only: it no longer calls `ensureMember` or writes `AccountMember` rows for migrated application members and instead collects unscoped app-only memberships for operator review in `OXY_ACCOUNTS_UNSCOPED_APP_MEMBERS` (file: `packages/api/scripts/migrate-accounts-40-app-members.ts`).
- Skip migration when an equivalent active `AccountMember` already exists by checking `AccountMember.exists(...)` so existing account members remain unaffected (file: `packages/api/scripts/migrate-accounts-40-app-members.ts`).
- Remove the legacy app-role → account-role mapper to avoid accidental reuse of the unsafe mapping (file: `packages/api/scripts/account-migration-lib.ts`).
- Add a regression test that ensures the migration source does not call `ensureMember`, does not write `AccountMember`, and emits the unresolved-members report (file: `packages/api/src/scripts/__tests__/accountAppMemberMigration.test.ts`).

### Testing

- Ran a dependency-free source guard script that scans the migration and confirmed it does not call `ensureMember` or write `AccountMember` and contains the `OXY_ACCOUNTS_UNSCOPED_APP_MEMBERS=` report, and that check passed. 
- Attempted to run the added Jest test but the environment lacked installed test tooling/dependencies so the suite could not be executed here (test runner not available). 
- Attempted TypeScript script typecheck which was blocked by missing type dependencies and a global TypeScript deprecation check in this environment, so a full `tsc` verification did not complete. 
- Performed repository static checks (diff/check) and the modified files pass local formatting/linting checks in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a719c80eedc83238dbd01aa0877b6ac)

### DIFF
--- a/packages/api/scripts/account-migration-lib.ts
+++ b/packages/api/scripts/account-migration-lib.ts
@@ -66,26 +66,6 @@ export function mapWorkspaceRole(role: string): AccountRole {
   }
 }
 
-/** Legacy ApplicationMember role → unified AccountRole on the owning account. */
-export function mapAppRole(role: string): AccountRole {
-  switch (role) {
-    case 'owner':
-      // An app owner becomes an admin of the owning account (account ownership
-      // is a stronger grant reserved for the account's true owner).
-      return 'admin';
-    case 'admin':
-      return 'admin';
-    case 'developer':
-      return 'developer';
-    case 'billing':
-      return 'billing';
-    case 'viewer':
-      return 'viewer';
-    default:
-      return 'viewer';
-  }
-}
-
 /** Connect to MongoDB using `MONGODB_URI`, or exit. */
 export async function connect(): Promise<void> {
   const uri = process.env.MONGODB_URI;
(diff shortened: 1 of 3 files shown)


NateIsern:
Closing: the target no longer exists. Both files this PR modifies — `packages/api/scripts/account-migration-lib.ts` and `packages/api/scripts/migrate-accounts-40-app-members.ts` — were deleted from `main` by `59830ee7d chore(api): delete the Mongo-only one-shot scripts (#889)`.

If application membership boundaries still need preserving in whatever replaced that migration path, that is a fresh patch against the current code, not a rebase of this one.

PR #936, opened 2026-08-10
fix(docker): freeze API image dependencies

NateIsern:
### Motivation
- Prevent the API Docker build from resolving mutable dependency versions at build time by restoring locked, frozen installs and pinning the Bun toolchain.
- Avoid materializing unrelated root/frontend dependencies into the production image while still validating the committed lockfile against the full workspace manifests.
- Reduce supply-chain risk where a future semver-compatible or malicious release could be pulled into a production image during Docker build.

### Description
- Add a `workspace-manifests` stage that copies the committed `package.json` and `bun.lock` alongside all workspace `package.json` files so Bun can validate the lockfile without copying unrelated source.
- Pin the Bun installer to `bun@1.3.14` in both builder and production stages to avoid package-manager drift.
- Use `bun install --frozen-lockfile` in the builder and `bun install --production --frozen-lockfile` in the runtime image, both scoped with `--filter` to materialize only the API dependency closure (`@oxyhq/contracts`, `@oxyhq/protocol`, `@oxyhq/federation`, `@oxyhq/core`, `@oxyhq/api`).
- Keep the existing build order and artifact copy steps (build contracts -> protocol -> core -> federation -> api; copy dist outputs and admin scripts), while ensuring the runtime image installs only frozen production deps.

### Testing
- Ran static assertions that check Bun is pinned twice, frozen installs are present twice, and workspace filters for the API closure appear; all assertions passed.
- Verified the manifest-only preparation prunes non-`package.json` files from `packages/` and that the manifest set contains more than the expected workspace manifests; the check passed.
- Performed repository diff checks for common issues (whitespace/fixme) which passed, and validated the Dockerfile changes via local textual inspections; Docker image builds were not executed because Docker is unavailable in the environment.
- Attempted to exercise the PR helper tooling but the local MCP helper runtime failed to start due to an unrelated environment Python module error, so PR metadata creation could not be completed through that helper (tooling failure, not a Dockerfile regression).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a719dd3b48c8323bbf1b58034b5db72)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,31 +9,32 @@
 ## Run:    docker run --env-file .env -p 8080:8080 oxy-api
 ##
 
-FROM node:20-alpine AS builder
-
-RUN npm install -g bun
+FROM alpine:3.22 AS workspace-manifests
 
 WORKDIR /app
 
-# Copy workspace root and override workspaces to only include api + core +
-# protocol + contracts + federation. `@oxyhq/api` depends on `@oxyhq/contracts` +
-# `@oxyhq/protocol` + `@oxyhq/federation` (workspace:*); core is retained for the
-# admin scripts that import packages/core/src/* at runtime (and core depends on
-# protocol).
-# Remove bun.lock since the workspace change invalidates it — bun will
-# resolve fresh dependencies (still deterministic from package.json versions).
-COPY package.json ./
-RUN node -e "const p=require('./package.json'); const catalog=p.workspaces?.catalog; const packages=['packages/contracts','packages/protocol','packages/federation','packages/core','packages/api']; p.workspaces=catalog?{packages,catalog}:packages; delete p.patchedDependencies; require('fs').writeFileSync('package.json', JSON.stringify(p, null, 2));"
+# Keep the committed monorepo lockfile and every workspace manifest together so
+# Bun can validate the lockfile, while excluding unrelated workspace source.
+COPY package.json bun.lock ./
+COPY packages/ packages/
+RUN find packages -type f ! -name package.json -delete
 
-# Copy package.json files for dependency resolution
-COPY packages/api/package.json packages/api/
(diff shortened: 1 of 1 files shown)


NateIsern:
Triaged against `main` @ `68d9d11c`. **Premise live, but the patch predates a substantial rewrite of this Dockerfile and needs reworking rather than rebasing.**

Still true: the image builds unlocked. `main` still carries

```dockerfile
# Remove bun.lock since the workspace change invalidates it — bun will
# resolve fresh dependencies (still deterministic from package.json versions).
...
# Install dependencies (no lockfile — workspace subset doesn't match the full monorepo lock)
RUN bun install
```

and "deterministic from package.json versions" is not what a caret range gives you — every rebuild re-resolves, so the image can drift from the reviewed tree without a single commit.

`main` has since added a compensating assertion right under it ("The install above is UNLOCKED, so this asserts the one property the lockfile would otherwise have guaranteed" — the single-copy express-types check). That narrows the blast radius to one known collision; it does not restore the pin.

Why this does not just rebase:

- `main`'s workspace list now includes `packages/mcp` and `packages/db`, with an explicit warning that a missing `workspace:*` entry is "not a degraded build, it is no build at all".
- `main` now also does `p.dependencies={}` to keep the Expo test app's root dependencies out of the server image. This PR keeps the real root `package.json`, so that exclusion has to be re-established (probably implied by `--filter`, but it needs to be shown, not assumed).
- The final stage on `main` has diverged from the version this diff edits.

The approach here is the right one — a `workspace-manifests` stage that carries every manifest so `--frozen-lockfile` can validate, with `--filter` to materialize only the API closure, and a pinned `bun@1.3.14`. It just needs to be written against today's file, with the Docker build job proving the image still comes out.

I am not rewriting a production image build unreviewed. Leaving open with the analysis.

